PR #17, opened 2018-08-18
Introduce "Train Phase Interface"

### LINKED ISSUE #15: Allow LRScheduler to step at the end of batches
TL;DR: some LRSchedulers must be able to update the learning rate (and other parameters) during the batches and not just at the end of an epoch. Should we implement this? What is the best way?

*Long version:*

I would love to see PyToune offer advanced learning rate schedules like 1 cycle policy!

PyTorch's `_LRScheduler` have a `step()` method which set the learning rate. The user is then responsible to call the `step()` method somewhere in the training loop. This normally happens at the end of each epoch, but it could also happen after each batch.

PyToune wraps PyTorch's `_LRScheduler`s via the `Callback` interface and calls the `step()` method `on_epoch_end`. This is great for convenience, but some LRSchedulers have to update the lr after each batch.

Here are a few options I see:
- We could update the `PyTorchLRSchedulerWrapper` and make it configurable when to trigger the `step()` call, and then use it something like this:
  ```python
  CosineAnnealingLR(step_on_batch_end=True, ...)
  ```

- We could introduce batch versions of all LRSchedulers.

- I use a simple callback `LRSetter` that gets a list of learning rates (and momentum values) and sets the lr (and momentum) on each batch end.

sotte:
This PR is going to introduce a fastai-like "train phase interface" and create the foundation for more complex and fine grained control over the training process.

The PR addresses and closes #15.

This PR adds the following
- Parameter spaces: `LinSpace` and `CosineSpace` lazily generate values in the specified domain.
- The `TrainPhase` class uses the parameter spaces and specifies how to configure the optimizer in each step. It return a dict of optimizer parameters.
- The `TrainPolicy` combines different `TrainPhase`s and updates the optimizer according to the parameters returned by the `TrainPolicy`s.

**TODO**
- [x] first implementation
- [x] tweak implementation
- [x] Store the state of the `TrainPolicy` and make it restartable
- [x] add "one cycle" policy
- [x] add "sgdr" policy
- [x] fully test everything (and switch from pytest back to UnitTest)
  - [x] spaces
  - [x] TrainPhase
  - [x] TrainPolicy
  - [x] integration test
- [x] docs
  - [x] write docstrings
  - [x] add to sphinx
  - [x] clean up sphinx warnings
- [x] add references to the papers
- [x] create a example notebook
- [x] clean up commit history

**Possible extensions**
- Different optimizers could be set for each phase
- More policies
- Set different learning rates for different part of the network

### DIFF
--- a/docs/rebuild_html_doc.sh
+++ b/docs/rebuild_html_doc.sh
@@ -1,1 +1,2 @@
+#!/bin/sh
 make clean && make html
--- a/docs/source/callbacks.rst
+++ b/docs/source/callbacks.rst
@@ -81,3 +81,25 @@ omitted.
 .. autoclass:: CosineAnnealingLR
 
 .. autoclass:: ReduceLROnPlateau
+
+Policies
+--------
+
+.. automodule:: pytoune.framework.callbacks.policies
+
+.. autoclass:: Phase
+
+.. autoclass:: OptimizerPolicy
+
+.. autofunction:: linspace
+
+.. autofunction:: cosinespace
+
+High Level Policies
+~~~~~~~~~~~~~~~~~~~
+
+Ready to use policies.
+
+.. autofunction:: one_cycle_phases
+
+.. autofunction:: sgdr_phases
--- a/examples/policy_cifar_example.ipynb
+++ b/examples/policy_cifar_example.ipynb
@@ -1,0 +1,245 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Train CIFAR with the `policy` module"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import torch"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "device = torch.device(\"cuda:0\" if torch.cuda.is_available() else \"cpu\")\n",
+    "device"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Load the data"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from torchvision import transforms\n",
+    "\n",
+    "_mean = [0.485, 0.456, 0.406]\n",
+    "_std = [0.229, 0.224, 0.225]\n",
+    "\n",
+    "\n",
+    "train_transform = transforms.Compose([\n",
+    "    transforms.RandomHorizontalFlip(),\n",
+    "    transforms.ColorJitter(.3, .3, .3),\n",
+    "    transforms.ToTensor(),\n",
+    "    transforms.Normalize(_mean, _std),\n",
+    "])\n",
+    "val_transform = transforms.Compose([\n",
+    "    transforms.ToTensor(),\n",
+    "    transforms.Normalize(_mean, _std),\n",
+    "])"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import torchvision.datasets as datasets\n",
+    "\n",
+    "root = \"data\"\n",
+    "train_ds = datasets.CIFAR10(root, train=True, transform=train_transform, download=True)\n",
+    "val_ds = datasets.CIFAR10(root, train=False, transform=val_transform, download=True)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from torch.utils.data import DataLoader\n",
+    "\n",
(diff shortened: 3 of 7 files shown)


sotte:
Current open questions:
- Do the spaces have to be a sequence (`__getitem__` and `__len__`) or would a simple iterable (`__iter__)` be sufficient?
- Does the Train phase have to be a sequence or is a iterable be sufficient?
- Specifying the steps in the `TrainPhase` and the spaces is repetitive and error prone. Can we avoid this?
- Should the `TrainPhase` have allow the user to specify the space in a shorter way like thi: `TrainPhase(100, lr=(0.1, 0.01))`
- Should the plot functions be part of the `TrainPhase` and `TrainPolicy`?

I'm tempted to turn the spaces and the TrainPhase into a simple iterable and just let the `TrainPolicy` track the state.

sotte:
- I switched the spaces and the `TrainPhase` to be in iterable and I like it. 
- The `TrainPhase` does not have a `step` keyword anymore. The length is defined by the spaces.
- The `TrainPolicy` has an `initial_step` parameter and should be restartable (must be tested).

sotte:
I forgot that you can't pickle generators. The spaces can't be generators.

sotte:
I think this is pretty feature complete. Feedback regarding the interface is very welcome :) The notebook gives a good overview.

If would start adding tests and write docs now when I get a LGTM from you @freud14.

freud14:
Hi,

Regarding

>I forgot that you can't pickle generators. The spaces can't be generators.

Do we need to be able to pickle the spaces? I mean, since we can put an `initial_step`, it is not really useful to pickle a TrainPhase or a TrainPolicy.

I would also put the code in the callback package instead of just the framework package.

Otherwise, it's fine by me. You can proceed.

Frédérik

freud14:
In the sphinx doc, it seems that there are bugs because some back-ticks are not properly closed/opened.

Comments for the notebooks:
- We should be able to use `from pytoune.framework import TrainPolicy` to import `TrainPolicy` for instance. 
- Since your predefined phases return lists, you should not need to do `[*sgdr_phases(base_cycle_length=200, repeat=3, cycle_mult=2)]`, just pass in `sgdr_phases(base_cycle_length=200, repeat=3, cycle_mult=2)`, for instance.

Hmm, I don't know for pytest. Is it possible to do classes with pytest? Or is it what you're trying to avoid? Is pytest the new way to do unit testing in python? I'm open to pytest, but I would like that it does not diverge too much from the style of the existing tests if you see what I mean.

Finally, there is a .pylintrc config file for PyLint in the root of the repo. Make sure that your code passes the linting test.

sotte:
Hey,
thanks for the feedback!

Regarding the notebook:
- The imports were already possible. I adjusted them to be of this format: `from pytoune.framework import TrainPolicy`
- You're totally right! I was using adding phases constantly that's why I used a list by default. Fixed that.
- I would add a bit more text to the notebooks once I'm done with everything.

I fixed the sphinx warnings (but there are still the `WARNING: Duplicate explicit target name: "pytorch".` from index.rst).

I'm passing the pylint checks now.

## PyTest
Regarding pytest: I've been using pytest for years for all my project and work and I simply prefer it.
- pytest's fixtures are super nice (much more powerful than `tearDown` and `setup`),
- you don't have to write a UnitTest class (but you can),
- you can simply use `assert` instead of creating a class, inhereting, using some form of `self.asser*`,
- the error msgs are nicer,
- you can parameterize your tests,
- pytest's testrunner also finds classic UnitTest classes and tests them,
- and more.

Than being said, you already have lots of tests in place and if you want to stick to that style I'll transform my tests accordingly.
(Unrelated to this project, check out pytest ;-) )

## Naming
The naming is not consistent. Sometimes I talk about policies, train policies, train phase api, etc. We should pick one term and stick to it. Maybe we don't need a big term like `TrainPolicy` but `Stepper` is sufficient. Also the module should be renamed accordingly. I don't have a strong preference but here is an idea:
```python
from pytoune.framework import train_phases

phases = train_phases.one_cycle_phases()
stepper = train_phases.Stepper(phases)

model.fit(..., callbacks=[stepper])
```

## Dependencies
The plot functions require matplotlib. Are you ok with adding matplotlib as dependency? Alternatively the plot functions could just print an error msg when matplotlib is not installed.

sotte:
I think I addressed all your points. Again, thanks for the feedback!

Ironically, PyLint seems to be a bit inconsistent with the `inconsistent-return-statement` warning :) It only shows the error for the first plot function, not the second plot function which has the same structure.
```
************* Module pytoune.framework.callbacks.policies
pytoune/framework/callbacks/policies.py:127:4: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)

------------------------------------------------------------------
Your code has been rated at 9.89/10 (previous run: 9.89/10, +0.00)
```
I just disabled the warning for the module.

I cleaned up the commit history and I think we can merge it soon. Anything you want me to improve?

freud14:
Alright, seems fine by me. I can merge the pull request?

sotte:
I'll reread the papers to double check the parameters for the two high level policies later today. Then we're good to go I think.

sotte:
Some tests, some docs, but otherwise it ok now. Feel free to merge this if it looks good to you.

freud14:
Merged. Thank you for your work and contribution.

sotte:
It was a lot of fun. I think I'll add some more features :)

PR #19, opened 2018-10-09
Group all tensorboard metrics in one experiment run

Ajk4:
Hi,

I believe current tensorboard callback does not use tensorboard summaries the way they were intended.

Currently different metrics groups are stored as different runs (left-hand side of tensorboard). 

Those are meant to represent different experiment run that you can compare against themselves. With current implementation this kind of bookkeeping is very difficult. 

This is current implementation:

old_way

With current approach is hard to compare different experiment and number of 'run' entries on left is overwhelming.

This PR makes TensorBoard callback behave as its counterpart in keras. All metrics are stored within one run only. That way you can easily compare different experiments.

See second screenshot for comparison:

tensorboard_1

Now I can easily see how many experiments I have and I can easily compare them against each other. If I want to select some plots subset on the right I can selectively hide/show any chart and/or use regex query. No need to clutter the 'runs' list!

### DIFF
--- a/pytoune/framework/callbacks/logger.py
+++ b/pytoune/framework/callbacks/logger.py
@@ -1,7 +1,6 @@
 import csv
 from .callbacks import Callback
 
-
 class Logger(Callback):
     def __init__(self, *, batch_granularity=False):
         super().__init__()
@@ -121,6 +120,7 @@ class TensorBoardLogger(Logger):
     """
     def __init__(self, writer):
         super().__init__(batch_granularity=False)
+
         self.writer = writer
 
     def _on_batch_end_write(self, batch, logs):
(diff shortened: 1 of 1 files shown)


freud14:
Hi,

Sorry for the delay. The thing is that I and people I know like to have the validation metrics in the same graphs as the training metrics since it allows us to compare for overfitting under the same scale instead of having two graphs per metric with maybe different scales. However, I agree with you that, with the way tensorboard is made, it is not very convenient if we want to compare between experiments. Since you seems to know tensorboard better than me, would you see an alternative?

Thank you.

Frédérik

Ajk4:
It's probably possible to have metrics and their validation counterparts on one charts. Hopefully it's possible with TensorboardX however I didn't research into it. I myself care about comparing multiple experiments at the moment.

What I can say with confidence though, is that this PR makes tensorboard callback compliant with its keras behaviour 😃

Ajk4:
I think I could do that in some future next time I fiddle with pytoune.

@freud14 

Also, if I would have managed to have val_X and X metrics on one charts, would you still want to have this feature flag?

freud14:
> Also, if I would have managed to have val_X and X metrics on one charts, would you still want to have this feature flag?

That would be the perfect solution but I don't know if it is at all possible with tensorboard.

> I think I could do that in some future next time I fiddle with pytoune.

Alright, let me know when it's done.

sotte:
My two cents: I use a custom TensorboardCallback exactly like the one from @Ajk4. I guess that is also the version most ppl would expect.

davebulaval:
@Ajk4 do you plan to finish this work soon ?

Ajk4:
Hi @davebulaval 

I am using my keras-like implementation and I'm happy with it. 

Currently I don't have plans recreating original poutyne behavior if that would you are asking for.

davebulaval:
> I think I could do that in some future next time I fiddle with pytoune.

I mean that, finishing the PR to be merge into master.

freud14:
Hi @Ajk4, for your information, I just introduced your suggested change with a flag in TensorBoardLogger: 93eda86e60b7cff4182a692b31850a7ce2f5a19f

PR #20, opened 2019-02-15
Update README.md

davidbelius:
The basic example as written produces an error with recently installed versions of python and pytorch (python 3.7.1 and pytorch 1.0.1), because the numpy *_y arrays are created as type int32 which get converted to torch IntTensors, while the cross-entropy loss of pytorch expects LongTensor. This fixes the error by creating the numpy arrays with type int64. See https://pste.eu/p/2N5f.html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,15 +38,16 @@ num_classes = 5
 
 num_train_samples = 800
 train_x = np.random.randn(num_train_samples, num_features).astype('float32')
-train_y = np.random.randint(num_classes, size=num_train_samples)
+train_y = np.random.randint(num_classes, size=num_train_samples).astype('int64')
+)
 
 num_valid_samples = 200
 valid_x = np.random.randn(num_valid_samples, num_features).astype('float32')
-valid_y = np.random.randint(num_classes, size=num_valid_samples)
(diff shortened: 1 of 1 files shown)


freud14:
Hi,
Your link doesn't work. Are you it is necessary? Because I just tried to setup a Python 3.7.1 with PyTorch '1.0.1.post2' and it seems to work fine.
Thank you.
Frédérik

davidbelius:
Hi,

Sorry about the link. Attached instead. With the latest Anaconda versions
of Python and pytorch the problem is there. Maybe it's an issue that only
affects pytorch 1.0.1 and was fixed in post2, I don't know!

Regards,
David

Den fre 15 feb. 2019 kl 15:46 skrev Frédérik Paradis <
notifications@github.com>:

freud14:
Hi,

I don't see any attachment. Are you on Windows? It seems that the problem is on Windows and not Linux. Anyway, could you also make a commit for the doc where the same exact code is present?

Thank you.

Fréédrik

davidbelius:
[int_long_issue.zip](https://github.com/GRAAL-Research/pytoune/files/2871157/int_long_issue.zip)

Second try with the attachments. Indeed I am on Windows 10.

PR #34, opened 2019-06-12
Formatting

TortillasAlfred:
Added [yapf](https://github.com/google/yapf) formatting to the whole repository.

The formatting is based on PEP8 with the addition of allowing lines to go up to the 120th character.

The pylint config has been updated correspondingly along with a small speedup due to multiprocessing.

### DIFF
--- a/poutyne/framework/callbacks/callbacks.py
+++ b/poutyne/framework/callbacks/callbacks.py
@@ -141,8 +141,8 @@ class Callback:
 
         Example::
 
-            logs = {'epoch': 6, 'time': 3.141519837, 'loss': 4.34462,
-                    'accuracy': 0.766, 'val_loss': 5.2352, 'val_accuracy': 0.682}
+            logs = {'epoch': 6, 'time': 3.141519837, 'loss': 4.34462, 'accuracy': 0.766,
+                    'val_loss': 5.2352, 'val_accuracy': 0.682}
         """
         pass
 
--- a/poutyne/framework/callbacks/callbacks.py
+++ b/poutyne/framework/callbacks/callbacks.py
@@ -110,6 +110,7 @@ class Callback:
         model (Model): a reference to the `Model` object which is using the
             callback.
     """
+
     def __init__(self):
         self.model = None
 
--- a/poutyne/framework/callbacks/callbacks.py
+++ b/poutyne/framework/callbacks/callbacks.py
@@ -1,6 +1,5 @@
 """
-The source code of this file was copied from the Keras project, and has been
-modified.
+The source code of this file was copied from the Keras project, and has been modified.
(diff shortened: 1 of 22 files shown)


poutyne/framework/callbacks/periodic.py
```diff
@@ -58,10 +58,10 @@ class PeriodicSaveCallback(Callback):
     `filename` is `weights.{epoch:02d}-{val_loss:.2f}.txt`, then `save_file()` will be called with a
     file descriptor for a file with the epoch number and the validation loss in the filename.
 
-    By default, the file are written atomically to the specified filename so that the training can
+    By default, the saves are written atomically to the specified filename so that the training can
```
freud14:
```suggestion
     By default, the file is written atomically to the specified filename so that the training can
```

poutyne/framework/callbacks/callbacks.py
```diff
@@ -141,17 +135,14 @@ def on_epoch_end(self, epoch, logs):
...
+                 * Other metrics: One key for each type of metrics. The metrics are also averaged.
+                 * val_loss': The average loss of the batches on the validation set.
+                 * Other metrics: One key for each type of metrics on the validation set. The metrics are also averaged.
 
         Example::
 
-            logs = {'epoch': 6, 'time': 3.141519837, 'loss': 4.34462, 'accuracy': 0.766,
-                    'val_loss': 5.2352, 'val_accuracy': 0.682}
+            logs = {'epoch': 6, 'time': 3.141519837, 'loss': 4.34462,
+                    'accuracy': 0.766, 'val_loss': 5.2352, 'val_accuracy': 0.682}
```
freud14: why is accuracy on the next line here?

TortillasAlfred: I moved it because it made the two lines more equal, but I guess having the val stats on their own line makes more sense.

TortillasAlfred: Fixed by d3b8dbbe0d45e523105f6a3c56b044bdaafcfd47.

poutyne/framework/callbacks/periodic.py
```diff
@@ -58,7 +58,7 @@ class PeriodicSaveCallback(Callback):
     `filename` is `weights.{epoch:02d}-{val_loss:.2f}.txt`, then `save_file()` will be called with a
     file descriptor for a file with the epoch number and the validation loss in the filename.
 
-    By default, the saves are written atomically to the specified filename so that the training can
+     By default, the file is written atomically to the specified filename so that the training can
```
freud14:
```suggestion
     By default, the file is written atomically to the specified filename so that the training can
```

PR #36, opened 2019-06-20
added yapf automatic lint when building

davebulaval:
Automation and validation of the yapf linter when building with travis.

Taken from [this](https://github.com/google/yapf/blob/master/.travis.yml) .yml file from yapl repo.


freud14:
It seems that the command you used reformate the files instead of just checking their formatting.

PR #39, opened 2019-08-19
Support for multiple batches per optimizer step

TortillasAlfred:
With this PR, one can now compute a running loss on multiple batches before proceeding with a weight upgrade, essentially allowing to train a NN on a batch size too big to fit within a GPU's memory by simply fitting multiple smaller batches before upgrading the weights.

Please note that, according to a given parameter `batches_per_step`, the current implementation will divide the running loss of all `batches_per_step` by `batches_per_step`, making the total loss used for the backpropagation an average of the losses on the desired previous batches. This only makes sense in case of losses which are already averaged along the batch size, instead of summed per example. However, since most losses are averaged by default over a given batch, I believe it makes sense to use this out of the box.

### DIFF
--- a/poutyne/framework/experiment.py
+++ b/poutyne/framework/experiment.py
@@ -299,9 +299,9 @@ class Experiment:
                                             epochs=epochs,
                                             steps_per_epoch=steps_per_epoch,
                                             validation_steps=validation_steps,
+                                            batches_per_step=batches_per_step,
                                             initial_epoch=initial_epoch,
-                                            callbacks=callbacks,
-                                            batches_per_step=batches_per_step)
+                                            callbacks=callbacks)
         finally:
             if tensorboard_writer is not None:
                 tensorboard_writer.close()
--- a/poutyne/framework/experiment.py
+++ b/poutyne/framework/experiment.py
@@ -239,8 +239,8 @@ class Experiment:
               epochs=1000,
               steps_per_epoch=None,
               validation_steps=None,
-              seed=42,
-              batches_per_step=1):
+              batches_per_step=1,
+              seed=42):
         # pylint: disable=too-many-locals
         if seed is not None:
             # Make training deterministic.
--- a/poutyne/framework/experiment.py
+++ b/poutyne/framework/experiment.py
@@ -241,6 +241,7 @@ class Experiment:
               validation_steps=None,
               seed=42,
               batches_between_backprops=1):
+        # pylint: disable=too-many-locals
         if seed is not None:
             # Make training deterministic.
             random.seed(seed)
--- a/poutyne/framework/experiment.py
+++ b/poutyne/framework/experiment.py
@@ -240,7 +240,7 @@ class Experiment:
               steps_per_epoch=None,
               validation_steps=None,
               seed=42,
-              batches_between_backprops=1):
+              batches_per_step=1):
         # pylint: disable=too-many-locals
         if seed is not None:
             # Make training deterministic.
@@ -301,7 +301,7 @@ class Experiment:
                                             validation_steps=validation_steps,
                                             initial_epoch=initial_epoch,
                                             callbacks=callbacks,
-                                            batches_between_backprops=batches_between_backprops)
+                                            batches_per_step=batches_per_step)
         finally:
             if tensorboard_writer is not None:
                 tensorboard_writer.close()
--- a/poutyne/framework/experiment.py
+++ b/poutyne/framework/experiment.py
@@ -239,7 +239,8 @@ class Experiment:
               epochs=1000,
               steps_per_epoch=None,
               validation_steps=None,
-              seed=42):
+              seed=42,
+              batches_between_backprops=1):
         if seed is not None:
             # Make training deterministic.
             random.seed(seed)
@@ -298,7 +299,8 @@ class Experiment:
                                             steps_per_epoch=steps_per_epoch,
                                             validation_steps=validation_steps,
                                             initial_epoch=initial_epoch,
-                                            callbacks=callbacks)
+                                            callbacks=callbacks,
+                                            batches_between_backprops=batches_between_backprops)
         finally:
             if tensorboard_writer is not None:
                 tensorboard_writer.close()
--- a/poutyne/framework/iterators.py
+++ b/poutyne/framework/iterators.py
@@ -53,7 +53,7 @@ class StepIterator:
         for step, data in _get_step_iterator(self.steps_per_epoch, self.generator):
             self.callback.on_batch_begin(step, {})
 
-            zero_all_gradients = ((step + 1) % self.steps_between_backprops == 0)
+            zero_all_gradients = (steps % self.steps_between_backprops == 1)
             do_backprop = (step % self.steps_between_backprops == 0) or (step == self.steps_per_epoch)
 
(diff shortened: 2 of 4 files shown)


poutyne/framework/experiment.py
```diff
@@ -239,7 +239,9 @@ def train(self,
               epochs=1000,
               steps_per_epoch=None,
               validation_steps=None,
-              seed=42):
+              seed=42,
+              batches_per_step=1):
```
davebulaval: Whe have a default value too for the constructor. We should only define one default.

poutyne/framework/iterators.py
```diff
@@ -27,11 +27,13 @@ def _get_step_iterator(steps, generator):
 
 
 class StepIterator:
-    def __init__(self, generator, steps_per_epoch, callback, metrics_names):
+    def __init__(self, generator, steps_per_epoch, callback, metrics_names, batches_per_step=1):
```
davebulaval: see comment before.

poutyne/framework/iterators.py
```diff
@@ -81,7 +83,8 @@ def __init__(self,
                  validation_steps,
                  initial_epoch=1,
                  callback,
-                 metrics_names):
+                 metrics_names,
+                 batches_per_step=1):
```
davebulaval: See comment before.

poutyne/framework/iterators.py
```diff
@@ -90,6 +93,7 @@ def __init__(self,
         self.initial_epoch = initial_epoch
         self.callback = callback
         self.metrics_names = metrics_names
+        self.batches_per_step = batches_per_step
```
davebulaval: Do we really need to keep this value ?

poutyne/framework/experiment.py
davebulaval: I think, the model should have the default value not the iterator.

poutyne/framework/model.py
```diff
@@ -334,17 +355,30 @@ def fit_generator(self,
...
+                   batches_per_step=1):
+        if zero_all_gradients:
+            self.optimizer.zero_grad()
 
         loss_tensor, metrics, pred_y = self._compute_loss_and_metrics(x,
                                                                       y,
                                                                       return_loss_tensor=True,
                                                                       return_pred=return_pred)
 
+        loss_tensor = loss_tensor / batches_per_step
```
davebulaval: do we need to validate a negative number ? or a too large batches per step value ?

poutyne/framework/experiment.py
TortillasAlfred: Not sure what you mean here: default value has been set to 1 for `batches_per_step` because the default use-case of the whole library is not what this particular PR aims to solve.

poutyne/framework/iterators.py
TortillasAlfred: Was a mistake. Fixed by 874c3f709364a90f3c5b28faf6118b27fd3185f2

poutyne/framework/iterators.py
TortillasAlfred: Also a mistake. Fixed by 874c3f709364a90f3c5b28faf6118b27fd3185f2

poutyne/framework/iterators.py
TortillasAlfred: Mistake. Fixed by 874c3f709364a90f3c5b28faf6118b27fd3185f2

poutyne/framework/model.py
TortillasAlfred: I know many libraries like Numpy and Scikit-learn do some kind of sanity check on all input arguments to provide their users with potentially more meaningful error messages. 

However, since this is not seen anywhere in Poutyne, I don't feel like it's needed to go beyond a clear description of the parameter's role, hence its possible values.

poutyne/framework/experiment.py
freud14: I think this is fine. However, maybe put the `batches_per_step` argument before the `seed` argument since the `seed` argument is less important and fits more with the previous arguments.

freud14: And do the same for every other places. I think we should put it after the `validation_steps` argument at every places.

poutyne/framework/model.py
```diff
@@ -334,17 +354,30 @@ def fit_generator(self,
...
```
freud14: I don't think that the loss that is sent in the logs should be divided. It should only be divided for the backward pass.

freud14: Also, if we want to support uneven batches, the loss should be multiplied by the batch size. Then, the gradient should be divided by the total number of examples for the optimizer step.

poutyne/framework/model.py
davebulaval: Good with that.

poutyne/framework/model.py
davebulaval: I think the same.

poutyne/framework/experiment.py
TortillasAlfred: Fixed by fecbe56534bf75d572f69f4c09a537505c4e8041

poutyne/framework/model.py
TortillasAlfred: Fixed by ec1d849831f6d02afc8c658098c0f5db20f4a65e.

poutyne/framework/experiment.py
```diff
@@ -298,7 +300,8 @@ def train(self,
                                             steps_per_epoch=steps_per_epoch,
                                             validation_steps=validation_steps,
                                             initial_epoch=initial_epoch,
-                                            callbacks=callbacks)
+                                            callbacks=callbacks,
+                                            batches_per_step=batches_per_step)
```
freud14: Put this after validation_steps.

poutyne/framework/iterators.py
```diff
@@ -28,6 +28,7 @@ def _get_step_iterator(steps, generator):
 
 class StepIterator:
     def __init__(self, generator, steps_per_epoch, callback, metrics_names):
+        # pylint: disable=too-many-arguments
```
freud14: Don't think it's necessary since pylint was passing before.

poutyne/framework/model.py
```diff
@@ -220,7 +225,8 @@ def fit(self,
                                   validation_steps=validation_steps,
                                   initial_epoch=initial_epoch,
                                   verbose=verbose,
-                                  callbacks=callbacks)
+                                  callbacks=callbacks,
+                                  batches_per_step=batches_per_step)
```
freud14: Put this after validation_steps.

poutyne/framework/model.py
```diff
@@ -323,32 +334,77 @@ def fit_generator(self,
...
+                    examples_in_step += step.size
+
+                    step.loss, step.metrics, _ = self._fit_batch(x,
+                                                                 y,
+                                                                 callback=callback_list,
+                                                                 step=step.number,
+                                                                 zero_all_gradients=zero_all_gradients,
+                                                                 do_backprop=do_backprop,
+                                                                 batch_size=step.size,
+                                                                 examples_in_step=examples_in_step)
```
freud14: I think that batch_size and examples_in_step should be None when batches_per_step is 1 so that the gradient and loss are not modified for nothing in that case.

PR #46, opened 2019-09-21
Inherit EpochMetrics from nn.Module and remove AllenNLP dependancies

freud14:
Inherit EpochMetrics from nn.Module and remove AllenNLP dependancies by copying the FBetaMeasure class and modifying it


poutyne/framework/metrics/epoch_metrics/__init__.py
```diff
@@ -0,0 +1,14 @@
+# pylint: disable=wildcard-import
+from .base import *
+from .fscores import *
+
+all_epochs_metrics_dict = dict(f1=F1)
+
+
+def get_epoch_metric(epoch_metric):
```
davebulaval: In the __init__, why here ? 

> If we apply this logic, we should do the same for batch metric

poutyne/framework/metrics/epoch_metrics/fscores.py
```diff
@@ -0,0 +1,259 @@
+"""
```
davebulaval: I'm really not sure of the plus value to remove dependancy from AllenNLP vs having a bigger code base to maintain.

> If we keep this code we should at least copy too the tests.

poutyne/framework/model.py
```diff
@@ -1007,7 +1007,7 @@ def _transfer_loss_and_metrics_modules_to_right_device(self):
         if isinstance(self.loss_function, torch.nn.Module):
             self.loss_function.to(self.device)
 
-        for metric in self.batch_metrics:
+        for metric in self.batch_metrics + self.epoch_metrics:
```
davebulaval: I prefer this approach as my previous one. Now it's called in the forward pass right ?

poutyne/framework/model.py
```diff
@@ -1007,7 +1007,7 @@ def _transfer_loss_and_metrics_modules_to_right_device(self):
...
             if isinstance(metric, torch.nn.Module):
```
davebulaval: Also this approach will allow to compute precision, recal and F1 score only with the FBeta right ? (What we previously spoke)

.pylintrc
```diff
@@ -136,7 +136,8 @@ disable=print-statement,
         duplicate-code,
         too-few-public-methods,
         unnecessary-pass,
-        cyclic-import
+        cyclic-import,
+        arguments-differ
```
davebulaval: Why are we skipping this now ?

poutyne/framework/metrics/epoch_metrics/fscores.py
davebulaval: We also should add the F1Score. Since most of the time we use the F metric with a beta of 1, it's convenient (also for the lambda user) to juste have a F1 score .

poutyne/framework/model.py
freud14: Yes.

.pylintrc
freud14: Because when you specify several arguments in the forward pass (instead of *args, **kwargs in the nn.Module), you get this warning. It was starting to be everywhere in the code.

poutyne/framework/metrics/epoch_metrics/fscores.py
freud14: It's already there.

poutyne/framework/model.py
freud14: I don't think that changes anything related to that.

poutyne/framework/metrics/epoch_metrics/fscores.py
freud14: The plus value is that, right now, we cannot add AllenNLP as a dependency since it is not compatible yet with PyTorch 1.2. So the people using Poutyne wanting to use the F1 score would be obligated to use PyTorch 1.1.

poutyne/framework/metrics/epoch_metrics/__init__.py
freud14: Because there other is some kind of cyclic import. But in the future, maybe we should use decorator instead of a dictionnary.

poutyne/framework/metrics/epoch_metrics/fscores.py
davebulaval: Ok so the plan is to add it, and leave it after ?

poutyne/framework/model.py
davebulaval: Nice.

poutyne/framework/model.py
davebulaval: Ok.

.pylintrc
davebulaval: Ok

poutyne/framework/metrics/epoch_metrics/fscores.py
davebulaval: Haven't see it, my bad.

poutyne/framework/metrics/epoch_metrics/__init__.py
davebulaval: Clearly a decorator is more appropriated.

poutyne/framework/metrics/epoch_metrics/fscores.py
freud14: I don't think having AllenNLP as a dependency is a good a idea. It comes itself with a lot of dependencies I would like to avoid. I think the project would be too linked with AllenNLP and we would be obligated to follow their development as well as PyTorch's. Also, the code for FBeta is not that huge and this will allow us to customize it more.

freud14: I just copied the tests from AllenNLP.

poutyne/framework/metrics/epoch_metrics/__init__.py
freud14: Yes, but we'll do that later.

poutyne/framework/metrics/epoch_metrics/fscores.py
TortillasAlfred: Agree with @freud14 here: Poutyne is basically a wrapper around Pytorch and should therefore only be strongly linked to Pytorch as a library. Any other big library dependence like the one we introduced with AllenNLP is more annoying than simply copy-pasting the code for further modifications if needed. The more libraries you strongly connect to, the more your library has to be constantly updated when those connected libraries are also updated.

davebulaval: Fair enough for the argument of preferring binding with torch and be minimal. 
I still think we open for copy-paste code base but for now it's good enough.

Issue #27, opened 2019-04-10
confused with poutyne.framework.Model and torch.nn.Module


menghuu:
I know `framework.Model` is:

>    The Model class encapsulates a PyTorch module/network, a PyTorch optimizer,
>    a loss function and metric functions. It allows the user to train a neural
>    network without hand-coding the epoch/step logic.

also, `nn.Module` is not `nn.Model`, but many people implement the `nn.Module` with the name `Model` and create instance with name `model`. It's little confusing when writing callback class (when accessing `self.model`) and the `framework.Model.__init__` has the parameter `model` which is in pytorch's concept.

Can we change the name of `Model` to `Engine` (just like [ignite](https://github.com/pytorch/ignite/blob/38a4f37de759e33bc08441bde99bcb50f3d81f55/ignite/engine/engine.py#L49) or [tnt](https://github.com/pytorch/tnt/blob/3ed904b2cbed16d3bab368bb9ca5adc876d3ce69/torchnet/engine/engine.py#L1)) or other name?

freud14:
Hi,
Sorry for the delay. Maybe we could rename the attribute model to something like "module" or "net/network"? What do you think?
Frédérik

menghuu:
two ways to make the concept clearly

1. change the `framework.Model` to `framework.Engine` or other
2. change the attribute to other name, just like @freud14 said

poutyne's easy-training/evaluation feature should be summary as:
> make a poutyne <1> containing a pytorch <2>, and <1> will train/fit and eval/test pytorch <2>

there are `Model`/`Engine`/`Module`/`Net`/`Network` to select

if poutyne's design aim is a full feature framework like keras+tensorflow, call <1> `Model` may be reasonable. 
in this condition, I think `net/network` should be better than `module` when refer the original pytorch `model`, sorry I am not a native English speaker, `module` just looks like `model` 😓 and normally pytorch `model` consist of many `modules`, `module` is not a good choice to describe the pytorch `model`. but, in other way, `module` object is consistent with class `nn.Module`, it hard to choice `module` or not. calling it `net/network` should not be confusing. (a shortcut is: can `net/network` cover all kinds of `models` in pytorch?)


if the aim is focusing on training or something like this, <1> should not be called with `Model`, many pytorch users have defined their `model/net/network` by pytorch, they only want a easy way to training their pytorch `model/net/network`.
[...]

PR #53, opened 2019-09-30
[NAMING] update model attribute

davebulaval:
changed attribute and param of model "model" to network to minimize the confusion from the model concept in pytorch.

Refer to issue #27 for discussion.


poutyne/framework/model.py
```diff
@@ -21,12 +21,12 @@
 
 class Model:
     """
-    The Model class encapsulates a PyTorch module/network, a PyTorch optimizer, a loss function and
+    The Model class encapsulates a PyTorch net/network, a PyTorch optimizer, a loss function and
```
TortillasAlfred:
```suggestion
    The Model class encapsulates a PyTorch network, a PyTorch optimizer, a loss function and
```

net is only a shortcut for network here.

poutyne/framework/model.py
```diff
@@ -136,18 +136,30 @@ class Model:
...
-    def __init__(self, model, optimizer, loss_function, *, metrics=None, batch_metrics=None, epoch_metrics=None):
+    def __init__(self,
+                 network,
+                 optimizer,
+                 loss_function,
+                 *,
+                 metrics=None,
+                 batch_metrics=None,
+                 epoch_metrics=None,
+                 model=None):
```
TortillasAlfred: With this current implementation, I can pass a model arg to my Model initialization, but it is undocumented and does absolutely nothing since it is not even assigned to self.network. I would definitely keep it in the args right now, if there is a model arg, first output a warning of future deprecation, then if there is also a non-null network, raise the invalid config error.

davebulaval: 1. `model = Model(pytorch_network, optimizer, ...)` then the pytorch_network will be pass to network. 

2.1 `model = Model(something_else, model=pytorch_network, optimizer = optimizer, ...)` depending of the case the something_else will be set to network, so the `if network and model is not None:` should catch this case.

2.2  `model = Model(model=pytorch_network, optimizer = optimizer, ...)`  will fail since the constructor is not correctly set. So Python will raise the appropriated error. (need to change is code)

I think we got all the edge case, but you are right, I forgot to add depreciation warning in the doc.

davebulaval: added doc in ebc7fc2

davebulaval:
Didnot put depreciation warning in experiment since beta.

poutyne/framework/model.py
freud14: In that case, since 2.1 and 2.2 crashes anyway, I don't see why we need the model argument. It won't be more backward compatible with or without it.

freud14:
Btw, there is a lot of `pytorch_module` to change in the tests.

davebulaval:
@freud14 true, i forgot about that part.

poutyne/framework/model.py
davebulaval: True. So I just raise a depreciation warning in all case and remove the `if network and model is not None:` test ?

freud14: I don't see where you could raise a deprecation warning. Just remove the `if network and model is not None:` test.

freud14: Oh if you meant the deprecation warning in the model property, yeah, sure, keep that one.

freud14:
Also, you forgot to modify the tutorial in the tutorials directory.

davebulaval:
@freud14 @TortillasAlfred just pushed the modification of all reference in Tests, README, examples and some forgotten reference in doc of model of `pytorch_module` to `pytorch_model`.

poutyne/framework/model.py
davebulaval: solved in d65545a

poutyne/framework/model.py
```diff
@@ -187,7 +195,7 @@ def fit(self,
             callbacks=None):
         # pylint: disable=line-too-long
         """
-        Trains the model on a dataset. This method creates generators and calls
+        Trains the network on a dataset. This method creates generators and calls
```
freud14: This is inconsistent with the doc of `fit_generator`.

davebulaval: solved in 3e0ae6f

PR #54, opened 2019-10-07
unnecessary code line in experiment

davebulaval:
This None affectation is unnecessary since `_init_tensorboard_callbacks` return a None if `SummaryWriter` is not present.


freud14:
If self.logging is false, it will crash in the finally block.

davebulaval:
True.

PR #55, opened 2019-10-07
Change naming of batch and epoch in callbacks

davebulaval:
I think the naming batch is misleading since it's a more a `batch_number` more than the usual `batch` concept.

### DIFF
--- a/poutyne/framework/callbacks/logger.py
+++ b/poutyne/framework/callbacks/logger.py
@@ -7,15 +7,15 @@ class Logger(Callback):
     def __init__(self, *, batch_granularity=False):
         super().__init__()
         self.batch_granularity = batch_granularity
-        self.epoch = 0
+        self.epoch_number = 0
 
     def on_train_begin(self, logs):
         metrics = ['loss'] + self.model.metrics_names
 
         if self.batch_granularity:
-            self.fieldnames = ['epoch', 'batch', 'size', 'time', 'lr']
+            self.fieldnames = ['epoch_number', 'batch', 'size', 'time', 'lr']
         else:
-            self.fieldnames = ['epoch', 'time', 'lr']
+            self.fieldnames = ['epoch_number', 'time', 'lr']
         self.fieldnames += metrics
         self.fieldnames += ['val_' + metric for metric in metrics]
         self._on_train_begin_write(logs)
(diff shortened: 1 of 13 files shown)


davebulaval:
After reading all the time batch and epoch was use in the doc, I think it would be more clear to use `batch_number` and `epoch_number` instead of verbally spoken name `step`.

davebulaval:
Also, as a discussion, I don't clearly see the reason of having a the `batch` argument if the logs contain the actual `batch_number`.

poutyne/framework/callbacks/logger.py
```diff
@@ -123,13 +124,16 @@ def __init__(self, writer):
         super().__init__(batch_granularity=False)
         self.writer = writer
 
-    def _on_batch_end_write(self, batch, logs):
+    def on_batch_end(self, batch_number, logs):
+        self._on_batch_end_write(batch_number, logs)
```
TortillasAlfred: Why add a on_batch_end() implementation here ?

TortillasAlfred:
> After reading all the time batch and epoch was use in the doc, I think it would be more clear to use `batch_number` and `epoch_number` instead of verbally spoken name `step`.

Not sure what you mean here, any example of what you would change ?

TortillasAlfred:
> Also, as a discussion, I don't clearly see the reason of having a the `batch` argument if the logs contain the actual `batch_number`.
> @freud14 @TortillasAlfred

I guess it is to ease reimplementation of ``on_batch_end()`` and other such methods. In the end, I do not see this as much of a big clean coding issue and the clarity gain, albeit small, is surely more impactful than the clean code violation.

poutyne/framework/callbacks/logger.py
davebulaval: This is a mistake. I was working on a implementation of a gradient tracker.

davebulaval:
> > After reading all the time batch and epoch was use in the doc, I think it would be more clear to use `batch_number` and `epoch_number` instead of as verbally spoken name `step`.
> 
> Not sure what you mean here, any example of what you would change ?

@freud14 and I spoke about the naming and we agreed on step, but as I was reading the doc, we were always referring as `batch: The batch number`. So I now disagreed with the name `step` for batch .

poutyne/framework/callbacks/logger.py
davebulaval: af3815a

freud14:
I'm not sure I see the issue with the current code and thus why this PR is necessary.

davebulaval:
Similar as PR #53 where naming seem confusing in the callback.

davebulaval:
It's just that the naming is misleading. Also added the close of the summary writer as similar of the  `CSVLogger`.

PR #57, opened 2019-10-11
Case of a sequence

davebulaval:
# Metric and loss in the specific case of a sequence

In the specific case of a sequence, there is minor change to be made to calculate some metric and the loss computation. Since, we have batch element with K tag prediction for N tokens, there is more dimension and the normal computation seem inacurate.

## F1 Sequence
A special case of F1 where we have a padding value needed to be mask. Maintly, the process is a flatten step before the forward pass and a test of condition of classes dimensions.

## Sequence Accuracy
In the case of the accuracy computation, we compute with a mask corresponding of the unpadded element.
> No tests

## Sequence Cross Entropy
Flatten before passing to the Cross Entropy.
> No tests

> Should the loss been in a module ?

### DIFF
--- a/poutyne/framework/loss.py
+++ b/poutyne/framework/loss.py
@@ -1,0 +1,30 @@
+from torch.nn import CrossEntropyLoss
+
+
+class SequenceCrossEntropy:
+    """
+    Wrapper around the cross entropy loss for a padded sequence
+    """
+
+    def __init__(self, padding_value=0):
+        self.loss_fct = CrossEntropyLoss(ignore_index=padding_value)
+
+    def __call__(self, y_pred, y):
+        """
+        Flatten the sequence before passing it to the CrossEntropyLoss.
+
+        Args:
+            y_pred (~torch.Tensor): The prediction from a model.
+            y (~torch.Tensor or Union[~torch.Tensor, ~torch.Tensor]): The target.
+                ~torch.Tensor if only the target.
+                Union[~torch.Tensor, ~torch.Tensor] if the target and a mask used to mask the padded elements.
+
+        Returns:
+                The loss value
+        """
+        if isinstance(y, tuple):
+            y = y[0]
+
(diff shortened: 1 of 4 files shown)


freud14:
I don't understand. If the class dimension in your output tensor is 1, everything should work fine with what we actually have?

davebulaval:
No it does not. When outputing a sequence with `[ [ [prediction vector token 1], [prediction vector token 2], ... ] ]` The accuracy is not calculated right and same for the loss. The main problem been for one example in the batch we have `N tokens`  of `k class` prediction.

davebulaval:
Error with mask in my example.

freud14:
```python
import torch
import torch.nn.functional as F
from poutyne.framework.metrics import acc

logits = torch.rand((32, 20, 10))
target = torch.randint(10, (32,20))
mask = torch.bernoulli(torch.ones((32, 20))*0.5)
# -100 is the default ignored index
masked_target = torch.where(mask.bool(), target, torch.ones_like(mask).long()*-100)          
masked_target_flat = masked_target.view(-1)
                                                                                                                                                                                          
probs = F.softmax(logits, dim=-1)                                                                                                                                                                                                    
probs_t = probs.transpose(-1, 1)                                                                                                                                                                                                     
logits_t = logits.transpose(-1, 1)                                                                                                                                                                                                   
probs_flat = probs.reshape(-1, 10)
logts_flat = logits.reshape(-1, 10)
print(F.cross_entropy(logits_t, masked_target))                                                                                                                                   
print(F.cross_entropy(logts_flat, masked_target_flat))                                                                                                                                                                      
print(acc(probs_t, masked_target))                                                                                                                                                                                                          
print(acc(probs.reshape(-1, 10), masked_target_flat))                                                                                                                                                                                   
```

This outputs:
> tensor(2.3608)
> tensor(2.3608)
> tensor(8.2237)
> tensor(8.2237)

PR #66, opened 2020-03-12
Tensorboard cleaner epoch logging

davebulaval:
- Remove logging of epoch number after every epoch end
- Added on train end the logging of the last epoch number

### DIFF
--- a/tests/framework/callbacks/test_logger.py
+++ b/tests/framework/callbacks/test_logger.py
@@ -45,6 +45,7 @@ class BaseCSVLoggerTest:
     CSVLogger = None
     batch_size = 20
     lr = 1e-3
+    num_epochs = 10
 
     def setUp(self):
         torch.manual_seed(42)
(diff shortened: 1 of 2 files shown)


tests/framework/callbacks/test_logger.py
```diff
@@ -151,6 +147,7 @@ def _test_logging(self, history):
             calls.append(call('loss', {'loss': h['loss'], 'val_loss': h['val_loss']}, h['epoch']))
             calls.append(call('lr', {'lr': 1e-3}, h['epoch']))
         self.writer.add_scalars.assert_has_calls(calls, any_order=True)
+        self.writer.add_scalars.assert_called_with('last-epoch', 10)
```
freud14: Use an argument or a constant for the 10.

davebulaval: done

PR #75, opened 2020-06-18
Update introduction_pytorch_poutyne.ipynb

nperraud:
Small typo


freud14:
Hi, 
Thank you for your PR. However, the file is indeed a CSV file. If you look at the instanciation of the CSVLogger callback, you can see that we pass `separator='\t'` to it. However, you're right that there is a missing "a". I just pushed the fix.
Thank you.

Issue #64, opened 2020-03-12
[FEATURE] colorlog 


davebulaval:
Use a similar approach as [Hydra](https://hydra.cc/docs/plugins/colorlog) to color the terminal log output.

- Use logger with a wrapper function
- Manage it like graal.utils

davebulaval:
https://pypi.org/project/colorama/

PR #76, opened 2020-07-05
Coloring the progress output for highlighting it

davebulaval:
(Ref #64)
Allow the user to trigger a default coloring scheme for the progress of the training.
Allow the user to specify a coloring scheme for the progress of the training.

### Example with the default coloring settings:

During training:
image

At the end of the training: (test is not colored)
image (1)


freud14:
We might want to be able to pass a subset of the keys so that the user can change some colors but not others.

freud14:
Also, it might be a good idea to add colorama in the [extra-requirements](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) and in the tests requirements.

davebulaval:
> We might want to be able to pass a subset of the keys so that the user can change some colors but not others.

I've thought about it and I don't have an opinion. If we do so, I will remove the validation of the user colors.
Do we print a warnings saying that we took some default colors ?

davebulaval:
> Also, it might be a good idea to add colorama in the [extra-requirements](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) and in the tests requirements.

I've also thought about it and was not sure if we wanted to have it in the dependencies. Is up to you.

davebulaval:
Will add to extra_requirements with message.
Will add in tests and keep skipif.
Will remove validate all args and override with user instead.

Issue #18, opened 2018-08-27
Progress bar


sudarshan85:
I think it would be nice to have a progress bar indicating the status of training. There's tons of examples to implement this, but I think the one used by FastAI would be nice. [Here](https://github.com/fastai/fast_progress/tree/master/fast_progress) is the repo for it and [here](https://github.com/fastai/fast_progress/tree/master/fast_progress) is the actual python file that implements it. I'd be happy to submit a PR for it, if someone can point me in the right direction of where this would go.

freud14:
Hi, the FastAI one seems to be especially for IPython Notebooks. As suggested in Issue #13, maybe it should be a callback that uses tqdm. Still, for IPython Notebooks, it could be interesting to have that. Either we integrate the fast_progress library into a callback or we code one directly in a callback.

davebulaval:
- fastprogress
- tqdm

davebulaval:
Note: _flag verbose for progresscallback (none, epoch, epoch and step), progresscallback only for every n epochs_

davebulaval:
Overhead:
fastprogress: unknow
tqdm: 60 ns
ProgressBar: 800 ns

PR #77, opened 2020-07-06
Progress bar feature

davebulaval:
(ref #18)

Added a progress bar with step/s for every epoch.
Added a total training progress bar with time per epoch at the end.

image
image (1)


poutyne/framework/callbacks/color_formatting.py
```diff
@@ -85,9 +104,14 @@ def on_epoch_end(self, epoch_number: int, epochs: int, epoch_total_time: float,
         Format on epoch end the epoch ratio (so far / to do), the total time for the epoch, the steps done and the
         metrics name and values.
         """
-        print(
-            self._epoch_formatting(epoch_number, epochs) + self._epoch_total_time_formatting(epoch_total_time) +
-            self._step_formatting(steps, steps) + self._metric_formatting(metrics_str) + Style.RESET_ALL)
+        sys.stdout.write(
```
davebulaval, line 107: Avec le print il y avait des sauts de lignes entre chaque epoch.

davebulaval:
Newer version without tqdm.
image (1)

freud14:
It seems that the last PR caused the step time to be displayed as an ETA when self.steps is None. However, it should be displayed as s/step as in https://github.com/GRAAL-Research/poutyne/blob/master/poutyne/framework/callbacks/progress.py#L44.

poutyne/framework/callbacks/progress_bar.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        percentage = f"{self.actual_steps / self.total_steps * 100:.2f}%"
+
+        progress_bar = self.progress_bar_formatting()
+
+        actual_time = time()
+        delta_t = actual_time - self.last_time
+        self.last_time = actual_time
+        self.mean_time = mean([self.mean_time, delta_t])
+        self.mean_rate = mean([round(1 / self.mean_time, 2), self.mean_rate])
```
freud14, line 79: I don't think we should use mean to do the average of only two values. Favor (x+y) / 2 instead. Also, I don't think the round makes much sense here. Finally, I think the mean_rate is already computed in ProgressionCallback (means_time). We should reuse it here.

davebulaval: 1. The reason for more clarity (mean seem more clear than (x+y)/2 (in my opinion).
2. True.
3. At first, I didn't want to pass another argument since they were 6. But with the removal of epoch at every epoch, I think it's okay now to pass it. Will change.

freud14: Maybe progress bar should not print the time but just the progress bar itself with percentages?

freud14: Since anyway, we do not compute the ETA in that class.

davebulaval: I agreed then

davebulaval:
Newest version without timer.
image

PR #95, opened 2020-11-19
Add progress bar for Test

davebulaval:
If we want to also add it for valid, we need to modify `callback` to have an `on_valid_begin` scheme and modify the `_validate` method of `model`.

Should I do it?


davebulaval:
To also add progress bar during val, I propose a mode = 'eval' [here](https://github.com/GRAAL-Research/poutyne/blob/7cd37a573c04d40db5897e696bc690f45b53edcc/poutyne/framework/iterators.py#L40).

And add on_predict_... to [callback](https://github.com/GRAAL-Research/poutyne/blob/7cd37a573c04d40db5897e696bc690f45b53edcc/poutyne/framework/callbacks/callbacks.py#L49).

poutyne/framework/callbacks/color_formatting.py
```diff
@@ -200,3 +203,28 @@ def _get_formatted_metrics(self, metrics_str: str) -> str:
             formatted_metrics += self.text_color + name + ":" + self.metric_value_color + value
 
         return formatted_metrics
+
+    def _batch_update(self,
+                      update,
```
freud14, line 208: I think that this argument argument is not necessary. This function should just return a new string and the += should be done by the caller.

poutyne/framework/callbacks/color_formatting.py
```diff
@@ -125,21 +125,24 @@ def on_train_batch_end(self,
...
 
-        update += self._get_formatted_metrics(metrics_str)
+    def on_test_batch_end(self,
+                          remaining_time: float,
+                          batch_number: int,
+                          metrics_str: str,
+                          steps: Union[int, None] = None) -> None:
+        # pylint: disable=too-many-arguments
+        """
+        Format on train batch end for a steps the epoch ratio (so far / to do), the total time for the epoch, the steps
```
freud14, line 140:
```suggestion
        Format on test batch end for a steps the epoch ratio (so far / to do), the total time, the steps
```

poutyne/framework/callbacks/progress.py
```diff
@@ -75,6 +75,36 @@ def on_train_batch_end(self, batch_number: int, logs: Dict):
...
+            self.color_progress.set_progress_bar(self.steps)
+
+    def on_test_batch_end(self, batch_number: int, logs: Dict) -> None:
+        if self.equal_weights:
+            self.step_times_weighted_sum += logs['time']
+            step_times_rate = self.step_times_weighted_sum / batch_number
+        else:
+            self.step_times_weighted_sum += batch_number * logs['time']
+            normalizing_factor = (batch_number * (batch_number + 1)) / 2
+            step_times_rate = self.step_times_weighted_sum / normalizing_factor
```
freud14, line 94: I think this should go in its own method since it seems repeated from `on_train_batch_end`.

poutyne/framework/iterators.py
```diff
@@ -44,6 +44,9 @@ def __init__(self, generator, steps_per_epoch, batch_metrics_names, callback=Non
             self.on_batch_begin = callback.on_test_batch_begin
             self.on_batch_end = callback.on_test_batch_end
 
+            callback.set_params({'steps': steps_per_epoch})
+            callback.on_test_begin({})
```
freud14, line 48: Why do you remove this from Model to put it here? I think we should just add the set_params call at the same place as on_test_begin in Model.

poutyne/framework/model.py
```diff
@@ -805,7 +805,16 @@ def predict_on_batch(self, x):
...
+    def evaluate(self,
+                 x,
+                 y,
+                 *,
+                 batch_size=32,
+                 return_pred=False,
+                 callbacks=None,
+                 verbose=True,
+                 progress_options: Union[dict, None] = None,
+                 **dataloader_kwargs):
```
freud14, line 817:
```suggestion
                 dataloader_kwargs=None):
```
I'm sorry, you're going to have to merge a conflict here since I just pushed this modification into dev.

davebulaval: gg not well played

poutyne/framework/callbacks/color_formatting.py
davebulaval: 8bc604e

poutyne/framework/callbacks/progress.py
davebulaval: 23d1cb5

poutyne/framework/iterators.py
davebulaval: 765a06d

poutyne/framework/model.py
```diff
@@ -890,6 +908,11 @@ def evaluate_dataset(self,
                 Used when using batched loading from a map-style dataset.
             dataloader_kwargs (dict, optional): Keyword arguments to pass to the PyTorch dataloaders created
                 internally.
+            verbose (bool): Whether to display the progress of the training.
```
freud14, line 911:
```suggestion
            verbose (bool): Whether to display the progress of the evaluation.
```

poutyne/framework/model.py
```diff
@@ -826,6 +835,11 @@ def evaluate(self, x, y, *, batch_size=32, return_pred=False, callbacks=None, da
                 testing. (Default value = None)
             dataloader_kwargs (dict, optional): Keyword arguments to pass to the PyTorch dataloaders created
                 internally.
+            verbose (bool): Whether to display the progress of the training.
```
freud14, line 838:
```suggestion
            verbose (bool): Whether to display the progress of the evaluation.
```

poutyne/framework/model.py
```diff
@@ -949,6 +976,11 @@ def evaluate_generator(self,
                 (Default value = False)
             concatenate_returns (bool, optional): Whether to concatenate the predictions
                 or the ground truths when returning them. (Default value = True)
+            verbose (bool): Whether to display the progress of the training.
```
freud14, line 979:
```suggestion
            verbose (bool): Whether to display the progress of the evaluation.
```

freud14:
A few additions that I think are necessary:
- [ ] At the end of testing, we should display the metrics averaged like at each end of epoch in addition to the epoch metrics.
- [ ] At the end of testing, we should display the total time of testing
- [ ] I don't know if it is desirable, maybe the metrics should be suffixed with `test_` in the same way that validation metrics are suffixed with `val_`?

freud14:
Epoch metric are missing at the end.
image

davebulaval:
> Epoch metric are missing at the end.
> image

Yeah.. with the fall back to the step iterator I forgot about that. I've pushed a fix for it 5ff92d8

PR #96, opened 2020-11-20
Add feature to load a checkpoint as a path and not only a int, best/last ckpt

davebulaval:
This feature allows a user to load a model as a `ckpt` path (as a `state_dict`).


tests/framework/experiment/experiment.py
```diff
@@ -0,0 +1,48 @@
...
+        self.pytorch_network = nn.Linear(1, 1)
+        self.optimizer = torch.optim.SGD(self.pytorch_network.parameters(), lr=1e-3)
+        self.loss_function = nn.MSELoss()
+
+        self.test_experiment = Experiment(test_checkpoints_path,
+                                          self.pytorch_network,
+                                          optimizer=self.optimizer,
+                                          loss_function=self.loss_function,
+                                          monitor_metric="loss",
+                                          monitor_mode="max")
```
freud14, line 26: Yeah, as we discussed, I think we could really just train for 1 epoch  in setUp. This adds a few lines for the creation of the temporary directory and removes the files for the checkpoints in the git repo.

PR #102, opened 2021-01-09
Notification callback

davebulaval:
Add a new callback to send notification to a channel using [Notif](https://notificationdoc.ca/index.html) notification logic.

### DIFF
--- a/extra_requirements.txt
+++ b/extra_requirements.txt
@@ -4,3 +4,4 @@ tensorboardX
 colorama
 torchvision
 pandas
+notif
--- a/extra_requirements.txt
+++ b/extra_requirements.txt
@@ -4,4 +4,3 @@ tensorboardX
 colorama
 torchvision
 pandas
-notif
--- a/poutyne/framework/callbacks/notification.py
+++ b/poutyne/framework/callbacks/notification.py
@@ -11,7 +11,7 @@ class Notificator(ABC):
     """
 
     @abstractmethod
-    def send_notification(self, message: str, subject: Union[str, None] = None) -> None:
+    def send_notification(self, message: str, *, subject: Union[str, None] = None) -> None:
         """
         Abstract method to send a notification.
 
--- a/poutyne/framework/callbacks/notification.py
+++ b/poutyne/framework/callbacks/notification.py
@@ -1,0 +1,31 @@
+from typing import Dict, Union
+
+from notif import Notificator
+
+from . import Callback
+
+
+class NotificationCallback(Callback):
+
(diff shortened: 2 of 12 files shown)


extra_requirements.txt
```diff
@@ -3,4 +3,5 @@ tensorboard
 tensorboardX
 colorama
 torchvision
-pandas
\ No newline at end of file
+pandas
+notif
```
freud14, line 7: I don't think we should add notif as an extra requirements. I think that the callback should specify the expected interface of the notificator and you could give your notif library as an example of library that implement the said interface.

poutyne/framework/callbacks/notification.py
```diff
@@ -0,0 +1,73 @@
+from typing import Dict, Union
```
freud14, line 1: You should import this this in `poutyne/framework/callbacks/__init__.py` so that we can do `from poutyne import NotificationCallback`.

davebulaval: Yes, I forgot.

extra_requirements.txt
davebulaval: I added an interface.

tests/framework/callbacks/test_notification.py
```diff
@@ -0,0 +1,137 @@
...
+                call_list.append(call.on_train_batch_end(step, {'batch': step, 'size': ANY, **train_batch_dict}))
+            call_list.append(call.on_epoch_end(epoch, logs[epoch - 1]))
+        call_list.append(call.on_train_end({}))
+
+        method_calls = self.notification_callback_mock.method_calls
+        self.assertIn(call.set_model(self.model), method_calls[:2])  # skip set_model and set param call
+        self.assertIn(call.set_params(params), method_calls[:2])
+
+        self.assertEqual(len(method_calls), len(call_list) + 2)  # for set_model and set param
+        self.assertEqual(method_calls[2:], call_list)
```
freud14, line 98: Not sure what is the point of this. This is tested at other places.

davebulaval: remove

poutyne/framework/callbacks/notification.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class Notificator(ABC):
+    """
+    The interface of the Notificator. It must at least implement a `send_notification` method.
+    The interface is similar to the `notif <https://notificationdoc.ca/index.html>`_ package.
+    """
+
+    @abstractmethod
+    def send_notification(self, message: str, subject: Union[str, None] = None) -> None:
```
freud14, line 14:
```suggestion
    def send_notification(self, message: str, *, subject: Union[str, None] = None) -> None:
```

PR #107, opened 2021-04-07
Restore best epoch at the end of training

davebulaval:
Added flag to restore the best epoch at the end of training in experiment.


freud14:
Hi, this already is the default behavior: https://github.com/GRAAL-Research/poutyne/blob/master/poutyne/framework/experiment.py#L379

PR #110, opened 2021-04-27
Example of classification and regression

mohammad-brdrn:


docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+
+    # Create required directories 
+    if not os.path.exists(data_root):
+    os.makedirs(data_root)
+    os.makedirs(dataset_folder)
+
+    # Download the dataset from the source
+    gdown.download(url, download_path, quiet=False)
+    gdown.download(land_mark_url, land_mark_path, quiet=False)
+    gdown.download(attr_url, attr_path, quiet=False)
```
freud14, line 90: You don't need to use gdown since the links are not on Google Drive.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+    attr_url = 'https://graal.ift.ulaval.ca/public/celeba/list_attr_celeba.txt'
+    land_mark_url = 'https://graal.ift.ulaval.ca/public/celeba/list_landmarks_align_celeba.txt'
+
+    # Path to download the dataset to
+    download_path = f'{data_root}/img_align_celeba.zip'
+    land_mark_path = f'{data_root}/list_landmarks_align_celeba.txt'
+    attr_path = f'{data_root}/list_attr_celeba.txt'
+
+    # Create required directories 
+    if not os.path.exists(data_root):
```
freud14, line 83: Use `exists_ok=True` with `os.makedirs` instead.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+    os.makedirs(dataset_folder)
+
+    # Download the dataset from the source
+    gdown.download(url, download_path, quiet=False)
+    gdown.download(land_mark_url, land_mark_path, quiet=False)
+    gdown.download(attr_url, attr_path, quiet=False)
+
+    # Unzip the downloaded file 
+    with zipfile.ZipFile(download_path, 'r') as ziphandler:
+    ziphandler.extractall(dataset_folder)
```
freud14, line 94: Missing indentation spaces.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+   
+Create a custom dataset class
+=============================
+
+As we are going to implement a multi-task problem by a single CNN, we should provide the CNN with the ground truth in a proper way. Here, we have two different tasks: classification and regression. In the classification task, the goal is to identify the gender. The labels of the gender for each image are saved in the `list_attr_celeba.txt` file, in which 1 stands for male and -1 for female. Since we consider the loss of both tasks simultaneously, we scale all target values to the range of [0,1]. Hence, the gender labels will be changed as well, 1 for male and 0 for female. For the localization part, the coordinates of the eyes (Left and Right) are provided in the `list_landmarks_align_celeba.txt` file. In addition to scaling the number to the range of [0,1], we also need to rescale the coordinates to the image's new size (224,224).
+
+.. code-block:: python
+
+    class CelebADataset(Dataset):
```
freud14, line 103: Why don't you use [torchvision's CelabA dataset](https://pytorch.org/vision/stable/datasets.html#celeba) instead?

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+
+    import math
+    import os
+    import matplotlib.pyplot as plt
+    import numpy as np
+    import pandas as pd
+    import gdown
+    import zipfile
+    import cv2
+    from natsort import natsorted
```
freud14, line 26: What packages do we need to install for these two?

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+    (x1, y1) = train_dataset[sample_number][1][1][0:2]
+    (x2, y2) = train_dataset[sample_number][1][1][2:4]
+    x1, x2 = int(x1*h), int(x2*h)
+    y1, y2 = int(y1*w), int(y2*w)
+    image_rgb	=	cv2.drawMarker(image_rgb, (x1,y1), (0,255,0))
+    image_rgb	=	cv2.drawMarker(image_rgb, (x2,y2), (0,255,0))
+    image_rgb = np.clip(image_rgb , 0, 1)
+    plt.imshow(image_rgb)
+    plt.axis('off')
+    plt.show()
```
freud14, line 184: The colors of the output are a little off.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+    image_rgb = cv2.cvtColor(np.float32(image), cv2.COLOR_BGR2RGB)
+    Gender = 'male' if int(train_dataset[sample_number][1][0])==1 else 'female'
+    print('Gender is: ', Gender)
+    w, h = train_dataset[sample_number][1][2]
+    (x1, y1) = train_dataset[sample_number][1][1][0:2]
+    (x2, y2) = train_dataset[sample_number][1][1][2:4]
+    x1, x2 = int(x1*h), int(x2*h)
+    y1, y2 = int(y1*w), int(y2*w)
+    image_rgb	=	cv2.drawMarker(image_rgb, (x1,y1), (0,255,0))
+    image_rgb	=	cv2.drawMarker(image_rgb, (x2,y2), (0,255,0))
```
freud14, line 180:
```suggestion
    image_rgb = cv2.drawMarker(image_rgb, (x1,y1), (0,255,0))
    image_rgb = cv2.drawMarker(image_rgb, (x2,y2), (0,255,0))
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+
+.. code-block:: python
+
+    class ClassifierLocalizer(nn.Module):
+        def __init__(self, model_name, num_classes=2):
+            super(ClassifierLocalizer, self).__init__()
+            self.num_classes = num_classes
+            
+            # create cnn model
+            model = models.__dict__[model_name](True)
```
freud14, line 201: Prefer using `getattr` instead of `__dict__`.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+    network = ClassifierLocalizer(model_name='resnet18')  # network without freezing any layer.
+
+Loss function
+=============
+
+As we discussed before, we have two different tasks in this example. These tasks need different loss functions; Cross-Entropy loss for the classification and Mean Square Error loss for the regression. Below, we define a new loss function class that sums both losses to considers them simultaneously. However, as the regression is relatively a harder task, we apply a higher weight to MSEloss.
+
+.. code-block:: python
+
+    class classification_regression_Loss(nn.Module):
```
freud14, line 239: Should be in Pascal case, that is `ClassificationRegressionLoss`

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+    (x1, y1) = predictions[1][sample_number][0:2]
+    (x2, y2) = predictions[1][sample_number][2:4]
+    x1, x2 = int(x1*h), int(x2*h)
+    y1, y2 = int(y1*w), int(y2*w)
+    image_rgb	=	cv2.drawMarker(image_rgb, (x1,y1), (0,255,0))
+    image_rgb	=	cv2.drawMarker(image_rgb, (x2,y2), (0,255,0))
+    image_rgb = np.clip(image_rgb , 0, 1)
+    plt.imshow(image_rgb)
+    plt.axis('off')
+    plt.show()
```
freud14, line 322: Same comments as above.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,324 @@
...
+Training
+========
+
+.. code-block:: python
+
+    optimizer = optim.Adam(network.parameters(), lr=0.0001, weight_decay=0)
+    loss_function = classification_regression_Loss()
+    #Step_Learning_Rate = StepLR(step_size=2 , gamma=0.1, last_epoch=-1, verbose=False)
+    exp = Experiment('./two_task_example', network, optimizer=optimizer, loss_function=loss_function, device="all")
+    exp.train(train_dataloader, valid_dataloader, callbacks=callbacks, epochs=num_epochs)
```
freud14, line 260: There is no `callbacks` defined above. Also, the ones you defined in the notebook are not useful since you are using Experiment and Experiment already has all of them.

freud14 (commented):
Hi,

I only looked at the code for now. I will wait for the corrections to look at the text afterwards.

You didn't include the images for the RST file.

Also, can you merge your branch with dev?

Thank you.

Frédérik

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: pip install natsort

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: I agree with you. however in the code:
**exp.train(train_dataloader, valid_dataloader, callbacks=callbacks, epochs=num_epochs)**
if i want to use the callbacks argument, i should define it first. isn't it?

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: i'll change it

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: i just found it easy for downloading. what do you recommend? 
how about "request.get" ?

docs/source/examples/classification_and_regression.rst
freud14: @mohammad-brdrn You can download the dataset like you do before instantiating the torchvision dataset. In that way, it won't try to download it from Internet.

docs/source/examples/classification_and_regression.rst
freud14: Yes, requests is good.

docs/source/examples/classification_and_regression.rst
freud14: I don't understand, what callbacks do you want to use?

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: you mean using torchvision.datasets.celebA(root=root, download=True)?
If so, that's the problem I mentioned. it always fails to download. also, torch communities also admit that and suggest downloading from the internet. for example link below:

https://github.com/pytorch/vision/issues/2262

freud14: Documentation says:

> download (bool, optional) – If true, downloads the dataset from the internet and puts it in root directory. If dataset is already downloaded, it is not downloaded again.

So if you download the dataset via the current links and put the files under the file hierarchy expected by torchvision.datasets.CelebA, it will work.

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: Solved

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: I put several days on this trying different strategies but still, we cannot use torch.datasets.CelbA as the standard dataset. because:

1)we publish this example mainly on google collab for the readers. Googlecolab does not keep downoladed files after runtime. so the problem still exists.
2) I tried downloading the file (first by setting download=true) and using it with (download=False). I tried it more than 9 days at almost any time (even midnight) to gives the same error.
3) I read almost all relevant posts on this issue and also inspected the root after the error. in fact, after the error (BadZipFile: File is not a zip file) i saw that it has downloaded all the files!! but it is not able to unzip the file! i tried downloading and unzipping the zipfile with other ways, but it doesnt work again.
4) can we proceed with the proposed dataset class? in meanwhile i will try other ways as well. i don't want to stop other parts for this long time, focusing on this part. maybe i can resume it along with other parts.

3afterzipls
1datadown
2zip

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: done

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: done

docs/source/examples/classification_and_regression.rst
freud14: Hi @mohammad-brdrn , sorry for the late response. Here is some code which works for using the CelebA dataset:

For loading the files:
```python
data_root = 'datasets'

base_url = "https://graal.ift.ulaval.ca/public/celeba/"

file_list = [
    "img_align_celeba.zip",
    "list_attr_celeba.txt", 
    "identity_CelebA.txt",
    "list_bbox_celeba.txt",
    "list_landmarks_align_celeba.txt",
    "list_eval_partition.txt"
]

# Path to folder with the dataset
dataset_folder = f'{data_root}/celeba'
os.makedirs(dataset_folder, exist_ok=True)

for file in file_list:
    url = f"{base_url}/{file}"
    wget.download(url, f"{dataset_folder}/{file}")

with zipfile.ZipFile(f"{dataset_folder}/img_align_celeba.zip", 'r') as ziphandler:
    ziphandler.extractall(dataset_folder)
```
Then, instantiating the datasets:
```python
train_dataset = datasets.CelebA(data_root, 
                                split='train', 
                                target_type=['attr', 'landmarks'], 
                                transform=tfms.ToTensor())
valid_dataset = datasets.CelebA(data_root, 
                                split='valid', 
                                target_type=['attr', 'landmarks'], 
                                transform=tfms.ToTensor())
test_dataset = datasets.CelebA(data_root, 
                               split='test', 
                               target_type=['attr', 'landmarks'], 
                               transform=tfms.ToTensor())
```

You had to look in the [code of CelebA](https://pytorch.org/vision/stable/_modules/torchvision/datasets/celeba.html#CelebA) to know how to do it.

freud14: Btw, when using wget, you should verify that the files do not exist before downloading. Otherwise, it creates files with a suffix "(1)".

docs/source/examples/classification_and_regression.rst
freud14: The text should specify that they are needed since they are not usually installed.

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: I applied the changes ( using torch.datasets.CelebA), While you check the overall format, I may work more on hyperparameters.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+
+Gender Classification and Eyes Location Detection: A Two Task Problem
+*********************************************************************
+
+.. note::
+
+    - See the notebook `here <https://github.com/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb>`_
+    - Run in `Google Colab <https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb>`_
+
+In this example, we are going to implement a multi-task problem. We try to identify the gender of the people, as well as locating their eyes in the image. Hence, we have two different tasks: classification (to identify the gender) and regression (to find the location of the eyes). We are going to use a single network (A CNN) to perform both tasks, however, we will need to apply different loss functions, each proper to a specific task.
```
freud14, line 12:
```suggestion
In this example, we are going to implement a multi-task problem. We try to identify the gender of the people, as well as locating their eyes in the image. Hence, we have two different tasks: classification (to identify the gender) and regression (to find the location of the eyes). We are going to use a single network (a CNN) to perform both tasks, however, we will need to apply different loss functions, each proper to a specific task.
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+    import numpy as np
+    import pandas as pd
+    import wget
+    import zipfile
+    import cv2
+    import torch
+    import torch.nn as nn
+    import torch.nn.functional as F
+    import torch.optim as optim
+    from torch.autograd import Variable
```
freud14, line 30: `Variable`s are deprecated and useless in recent versions of PyTorch. Thus, you can remove its uses everywhere and everything will work as expected.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+    ]
+
+    # Path to folder with the dataset
+    dataset_folder = f'{data_root}/celeba'
+    os.makedirs(dataset_folder, exist_ok=True)
+
+    for file in file_list:
+        url = f"{base_url}/{file}"
+        if not os.path.exists(f"{dataset_folder}/{file}"):
+        wget.download(url, f"{dataset_folder}/{file}")
```
freud14, line 87:
```suggestion
        if not os.path.exists(f"{dataset_folder}/{file}"):
            wget.download(url, f"{dataset_folder}/{file}")
```

Missing tab.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+   
+Now, as the dataset id downloaded, we can define our datasets and dataloaders in its original way.
+
+.. code-block:: python
+
+    transforms = tfms.Compose ([
+                            tfms.Resize((image_size, image_size)),
+                            tfms.ToTensor(),
+                            tfms.Normalize(imagenet_mean, imagenet_std)
+    ])
```
freud14, line 100:
```suggestion
    transforms = tfms.Compose([
        tfms.Resize((image_size, image_size)),
        tfms.ToTensor(),
        tfms.Normalize(imagenet_mean, imagenet_std)
    ])
```
Too many spaces.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+    print (train_dataset[0])
+
+Regarding the complexity of the problem and the high number training/valid samples, we can seperate and use a portion of data as below:
+
+.. code-block:: python
+
+    train_subset = Subset(train_dataset, np.arange(1,10000))
+    valid_subset = Subset(valid_dataset, np.arange(1,2000))
+    train_dataloader = DataLoader(train_subset, batch_size=batch_size, shuffle=True)
+    valid_dataloader = DataLoader(valid_subset, batch_size=batch_size, shuffle=False)
```
freud14, line 131: Not sure why we take a subset of the dataset. Is it because full training is too long?

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+Here, we can see an example from the training dataset. It shows an image of a person, printing the gender and also showing the location of the eyes. It is worth mentioning that as we resize the image, the coordinates of the eyes should also be changed with same ratio.
+
+.. code-block:: python
+
+    sample_number = 189
+    image = train_dataset[sample_number][0]
+    image = image.permute(1,2,0).detach().numpy()
+    image_rgb = cv2.cvtColor(np.float32(image), cv2.COLOR_BGR2RGB)
+    image_rgb = image_rgb * imagenet_std + imagenet_mean
+    Gender = 'male' if int(train_dataset[sample_number][1][0][20])==1 else 'female'
```
freud14, line 142: I'm guessing that 20 is the index of the gender. I think it should be defined as a constant and explained after its definition. Otherwise, one might be confused by the code.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+    print('Gender is: ', Gender)
```
freud14, line 143:
```suggestion
    gender = 'male' if int(train_dataset[sample_number][1][0][20])==1 else 'female'
    print('Gender is: ', gender)
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+
+.. code-block:: python
+
+    sample_number = 35
+    image = train_dataset[sample_number][0]
+    image = image.permute(1,2,0).detach().numpy()
+    image_rgb = cv2.cvtColor(np.float32(image), cv2.COLOR_BGR2RGB)
+    image_rgb = image_rgb * imagenet_std + imagenet_mean
+    Gender = 'male' if np.argmax(predictions[0][sample_number])==0 else 'female'
+    print('Gender is: ', Gender)
```
freud14, line 289:
```suggestion
    gender = 'male' if np.argmax(predictions[0][sample_number])==0 else 'female'
    print('Gender is: ', gender)
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+    plt.imshow(image_rgb)
+    plt.axis('off')
+    plt.show()
+
+.. image:: /_static/img/classification_and_regression/dataset_sample.png
+
+Network
+=======
+
+Below, we define a new class, named 'ClassifierLocalizer, which accepts a pre-trained CNN and changes its last fully connected layer to be proper for the two task problem. The new fully connected layer contains 6 neurons, 2 for the classification task (male or female) and 4 for the localization task (x and y for the left and right eyes). Moreover, to put the location results on the same scale as the class scores, we apply the sigmoid function to the neurons assigned for the localization task.
```
freud14, line 166:
```suggestion
Below, we define a new class, named `ClassifierLocalizer`, which accepts a pre-trained CNN and changes its last fully connected layer to be proper for the two task problem. The new fully connected layer contains 6 neurons, 2 for the classification task (male or female) and 4 for the localization task (x and y for the left and right eyes). Moreover, to put the location results on the same scale as the class scores, we apply the sigmoid function to the neurons assigned for the localization task.
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+            # remove fc layers and add a new fc layer
+            num_features = model.fc.in_features
+            model.fc = nn.Linear(num_features, 6) # classifier + localizer
+            self.model = model
+        
+        def forward(self, x):
+            x = self.model(x)                    # extract features from CNN
+            scores = x[:, :self.num_classes]     # class scores
+            coords = x[:, self.num_classes:]     # coordinates
+            return [scores, torch.sigmoid(coords)]   # sigmoid output is in the range of [0, 1]
```
freud14, line 187:
```suggestion
            return scores, torch.sigmoid(coords)   # sigmoid output is in the range of [0, 1]
```
Not need to return a list, a simple tuple will do.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+
+Regarding the complexity of the problem, the number of the samples in the training dataset, and the similarity of the training dataset to the ImageNet dataset, we may decide to freeze some of the layers. In our current example, based on the mentioned factors, we freeze just the last fully connected layer.
```
freud14, line 189:
```suggestion
Regarding the complexity of the problem, the number of the samples in the training dataset, and the similarity of the training dataset to the ImageNet dataset, we may decide to freeze some of the layers. In our current example, based on the mentioned factors, we freeze all layers but the last fully connected layer.
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+        for name, param in network.named_parameters():
+            if not name.startswith('fc.'):
+                param.requires_grad = False
+
+    freeze_weights(network)
+    print(network)
+
+.. code-block:: python
+
+    network = ClassifierLocalizer(model_name='resnet18')  # network without freezing any layer.
```
freud14, line 205: Not sure why you freeze the layers if you create after with no frozen layers.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+
+Loss function
+=============
+
+As we discussed before, we have two different tasks in this example. These tasks need different loss functions; Cross-Entropy loss for the classification and Mean Square Error loss for the regression. Below, we define a new loss function class that sums both losses to considers them simultaneously. However, as the regression is relatively a harder task, we apply a higher weight to MSEloss.
```
freud14, line 210: > we apply a higher weight to MSEloss.

Not sure where you apply this weight in the code below.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+            super(ClassificationRegressionLoss, self).__init__()
+            self.ce_loss = nn.CrossEntropyLoss() # size_average=False
+            self.mse_loss = nn.MSELoss()
+            
+        def forward(self, y_pred, y_true):
+            loss_cls = self.ce_loss(y_pred[0], Variable(y_true[0][:,20])) # Cross Entropy Error (for classification)
+            loss_reg1 = self.mse_loss(y_pred[1][:,0], Variable(y_true[1][:,0]/h, requires_grad=True)) # Mean Squared Error for X_L
+            loss_reg2 = self.mse_loss(y_pred[1][:,1], Variable(y_true[1][:,1]/w, requires_grad=True)) # Mean Squared Error for Y_L
+            loss_reg3 = self.mse_loss(y_pred[1][:,2], Variable(y_true[1][:,2]/h, requires_grad=True)) # Mean Squared Error for X_R
+            loss_reg4 = self.mse_loss(y_pred[1][:,3], Variable(y_true[1][:,3]/w, requires_grad=True)) # Mean Squared Error for Y_R
```
freud14, line 225: This can be done in a single line, such as `loss_reg = self.mse_loss(y_pred[1], y_true[1])`.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+    optimizer = optim.Adam(network.parameters(), lr=learning_rate, weight_decay=0)
+    loss_function = ClassificationRegressionLoss()
+    #Step_Learning_Rate = StepLR(step_size=2 , gamma=0.1, last_epoch=-1, verbose=False)
+    exp = Experiment('./two_task_example', network, optimizer=optimizer, loss_function=loss_function, device="all")
+    exp.train(train_dataloader, valid_dataloader, callbacks=callbacks, epochs=num_epochs)
+
+Evaluation
+==========
+
+As you have also noticed from the training logs, we have achieved the best performance (considering the validation loss) at the 15th epoch. The weights of the network for the corresponding epoch have been automatically saved and we use these parameters to evaluate our algorithm visually. Hence,  we take advantage of evaluate function of Poutyne, and apply it to the validation dataset. It provides us the predictions as well as the ground-truth for comparison, in case of need.
```
freud14, line 243: > we have achieved the best performance (considering the validation loss) at the 15th epoch.

Is this always the case? Anyway, I think you could the load_checkpoint function in Experiment that allows to load the best checkpooint.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+    set_seeds(42)
+    imagenet_mean = [0.485, 0.456, 0.406]  # mean of the ImageNet dataset for normalizing 
+    imagenet_std = [0.229, 0.224, 0.225]  # std of the ImageNet dataset for normalizing
+    device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
+    print('The running processor is...', device) 
+
+CelebA Dataset
+==============
+
+The section below consists of a few lines of codes that help us download celebA dataset from a public web source and unzip them. Downloading the Celeba dataset can be also done directly by `torch.datasets.CelebA(data_root, download =  True)`. however, due to the high traffic of Google Drive ( the main source of the dataset) it usually fails to function. hence we decided to download it from another public source but use it with `torch.datasets.CelebA()`.
```
freud14, line 59:
```suggestion
The section below consists of a few lines of codes that help us download CelebA dataset from a public web source and unzip them. Downloading the CelebA dataset can be also done directly by `torch.datasets.CelebA(data_root, download=True)`. However, due to the high traffic on the dataset's Google Drive (the main source of the dataset), it usually fails to function. Hence we decided to download it from another public source but use it with `torch.datasets.CelebA()`.
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,302 @@
...
+
+Fetching data 
+=============
+
+The section below consists of a few lines of codes that help us download celebA dataset from a public web source and unzip them.
```
freud14, line 64: This repeats what has just been said.

docs/source/examples/classification_and_regression.rst
freud14: This was not done in the last commit.

mohammad-brdrn: I added a text in the IPYNP as I included the codes to install them. in previous examples, you mentioned removing installation codes ( pip install poutyne for example) from the rest file. 
So in the rest file do you want me to just mention the need for installation of also to include the installation codes?

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: Sure. I used it o activate the gradient ( requires_grad=True). So is it also automatically done in the latest version?

docs/source/examples/classification_and_regression.rst
freud14: Just for these since they are unusual, add a line in the RST saying you use these libraries and add the lines to install them.

docs/source/examples/classification_and_regression.rst
freud14: In the places you used `requires_grad=True`, you don't need to.

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: There is a huge number of images with a limited variation. (for example, eyes location don't vary very much). so using all images (189,643 images) is just time-consuming.

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: I will remove that, I was just examining it as a hyper parameter

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: Yes if we just normalize the coordinates to another constant number. here I normalized them with their related coordinate length (h or w).

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: I didn't get your point. Do you mean I have repeated it in another place?

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: No, i will change it to clearly say that in this run (for example) it happened in the 15th epoch and the Poutyne automatically saves the best weights (whether it is in 15 to 14, or...). I also used evaluate function to access predictions and expectations simultaneously.

docs/source/examples/classification_and_regression.rst
freud14: Alright, maybe note that in the text.

docs/source/examples/classification_and_regression.rst
freud14: Oh right.

docs/source/examples/classification_and_regression.rst
freud14: I'm not sure it's the right thing to do because if someone run the notebook, it will not work and will have to manually change the filename in the code.

docs/source/examples/classification_and_regression.rst
freud14: Yeah, the text above also says that.

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: I changed it to select checkpoint which contains the best weights without mentioning the epoch number directly. i think this file contains the best weights, doesn't it?

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,303 @@
...
+            self.ce_loss = nn.CrossEntropyLoss() # size_average=False
+            self.mse_loss = nn.MSELoss()
+            
+        def forward(self, y_pred, y_true):
+            loss_cls = self.ce_loss(y_pred[0], Variable(y_true[0][:,20])) # Cross Entropy Error (for classification)
+            loss_reg1 = self.mse_loss(y_pred[1][:,0], y_true[1][:,0]/h) # Mean Squared Error for X_L
+            loss_reg2 = self.mse_loss(y_pred[1][:,1], y_true[1][:,1]/w) # Mean Squared Error for Y_L
+            loss_reg3 = self.mse_loss(y_pred[1][:,2], y_true[1][:,2]/h) # Mean Squared Error for X_R
+            loss_reg4 = self.mse_loss(y_pred[1][:,3], y_true[1][:,3]/w) # Mean Squared Error for Y_R
+            total_loss = loss_cls + W * (loss_reg1 + loss_reg2 + loss_reg3 + loss_reg4)
```
freud14, line 227: `W` should be parameter of the `__init__` method of the class, not a global as currently.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,303 @@
...
+   
+Now, as the dataset id downloaded, we can define our datasets and dataloaders in its original way.
+
+.. code-block:: python
+
+    transforms = tfms.Compose ([
+        tfms.Resize((image_size, image_size)),
+        tfms.ToTensor(),
+        tfms.Normalize(imagenet_mean, imagenet_std)
+        ])
```
freud14, line 105:
```suggestion
    transforms = tfms.Compose([
        tfms.Resize((image_size, image_size)),
        tfms.ToTensor(),
        tfms.Normalize(imagenet_mean, imagenet_std)
    ])
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,303 @@
...
+        target_type=['attr', 'landmarks'],
+        transform=transforms)
+    valid_dataset = datasets.CelebA(data_root, 
+        split='valid', 
+        target_type=['attr', 'landmarks'], 
+        transform=transforms)
+    test_dataset = datasets.CelebA(data_root, 
+        split='test', 
+        target_type=['attr', 'landmarks'], 
+        transform=transforms)
```
freud14, line 117:
```suggestion
    train_dataset = datasets.CelebA(
        data_root,
        split='train',
        target_type=['attr', 'landmarks'],
        transform=transforms
    )
    valid_dataset = datasets.CelebA(
        data_root, 
        split='valid', 
        target_type=['attr', 'landmarks'], 
        transform=transforms)
    test_dataset = datasets.CelebA(
        data_root, 
        split='test', 
        target_type=['attr', 'landmarks'], 
        transform=transforms
    )
```

docs/source/examples/classification_and_regression.rst
freud14: checkpoint.ckpt does not contain the best checkpoint but the last checkpoint. You should really use the load_checkpoint method in Experiment since it is there for that. You can pass it `'best'` to load the best checkpoint. Also, you can use the test method in Experiment instead of the evaluate_generator in Model to do the same thing you're currently doing. Or, you can get the Model inside the Experiment via `expt.model`.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,308 @@
...
+Evaluation
+==========
+
+As you have also noticed from the training logs, in this try we achieved the best performance (considering the validation loss) at the 15th epoch. The weights of the network for the corresponding epoch have been automatically saved by the `Experiment` function and we use these parameters to evaluate our algorithm visually. For this purpose, we utulize the load_checkpoint method and set its argument to `best` to load the best weights of the model automatically. Finally,  we take advantage of the `evaluate` function of Poutyne, and apply it to the validation dataset. It provides us the predictions as well as the ground-truth for comparison, in case of need.
+
+.. code-block:: python
+
+    exp.load_checkpoint('best')
+    model = exp.model
+    loss, predictions, Ground_Truth = model.evaluate_generator(valid_dataloader, return_pred=True, return_ground_truth=True)
```
freud14, line 255:
```suggestion
    loss, predictions, ground_truth = model.evaluate_generator(valid_dataloader, return_pred=True, return_ground_truth=True)
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,312 @@
...
+            total_loss = loss_cls + self.W * (loss_reg1 + loss_reg2 + loss_reg3 + loss_reg4)
+            return total_loss
+
+Training
+========
+
+.. code-block:: python
+
+    optimizer = optim.Adam(network.parameters(), lr=learning_rate, weight_decay=0)
+    loss_function = ClassificationRegressionLoss()
```
freud14, line 245:
```suggestion
    loss_function = ClassificationRegressionLoss(W)
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+        "colab": {
+          "base_uri": "https://localhost:8080/"
+        },
+        "id": "VAG5NDFCiqxG",
+        "outputId": "08544c29-1077-4f2d-94d4-51bd37cfd11a"
+      },
+      "source": [
+        "%pip install poutyne          # to install the Poutyne library\n",
+        "%pip install wget             # to install the wget library in order to download data\n",
+        "#%pip install opencv-python    # to install the cv2 (opencv) library"
```
freud14, line 296:
```suggestion
        "%pip install opencv-python    # to install the cv2 (opencv) library"
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+        "]\n",
+        "\n",
+        "# Path to folder with the dataset\n",
+        "dataset_folder = f'{data_root}/celeba'\n",
+        "os.makedirs(dataset_folder, exist_ok=True)\n",
+        "\n",
+        "for file in file_list:\n",
+        "    url = f\"{base_url}/{file}\"\n",
+        "    if not os.path.exists(f\"{dataset_folder}/{file}\"):\n",
+        "      wget.download(url, f\"{dataset_folder}/{file}\")\n",
```
freud14, line 475:
```suggestion
        "        wget.download(url, f\"{dataset_folder}/{file}\")\n",
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+        "      super(ClassifierLocalizer, self).__init__()\n",
+        "      self.num_classes = num_classes\n",
+        "      \n",
+        "      # create cnn model\n",
+        "      model = getattr(models, model_name)(pretrained=True)\n",
+        "      \n",
+        "      # remove fc layers and add a new fc layer\n",
+        "      num_features = model.fc.in_features\n",
+        "      model.fc = nn.Linear(num_features, 6) # classifier + localizer\n",
+        "      self.model = model\n",
```
freud14, line 711:
```suggestion
        "    def __init__(self, model_name, num_classes=2):\n",
        "        super(ClassifierLocalizer, self).__init__()\n",
        "        self.num_classes = num_classes\n",
        "        \n",
        "        # create cnn model\n",
        "        model = getattr(models, model_name)(pretrained=True)\n",
        "\n",
        "        # remove fc layers and add a new fc layer\n",
        "        num_features = model.fc.in_features\n",
        "        model.fc = nn.Linear(num_features, 6) # classifier + localizer\n",
        "        self.model = model\n",
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+        "image_rgb = cv2.cvtColor(np.float32(image), cv2.COLOR_BGR2RGB)\n",
+        "image_rgb = image_rgb * imagenet_std + imagenet_mean\n",
+        "gender = 'male' if np.argmax(predictions[0][sample_number])==0 else 'female'\n",
+        "print('Gender is: ', gender)\n",
+        "(x_L, y_L) = predictions[1][sample_number][0:2]*image_size\n",
+        "(x_R, y_R) = predictions[1][sample_number][2:4]*image_size\n",
+        "x_L, x_R = int(x_L), int(x_R)\n",
+        "y_L, y_R = int(y_L), int(y_R)\n",
+        "image_rgb\t= cv2.drawMarker(image_rgb, (x_L,y_L), (0,255,0))\n",
+        "image_rgb\t= cv2.drawMarker(image_rgb, (x_R,y_R), (0,255,0))\n",
```
freud14, line 1298:
```suggestion
        "image_rgb = cv2.drawMarker(image_rgb, (x_L,y_L), (0,255,0))\n",
        "image_rgb = cv2.drawMarker(image_rgb, (x_R,y_R), (0,255,0))\n",
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+        "(x_L, y_L) = train_dataset[sample_number][1][1][0:2]  # The coordinates vector of the datasets starts with X_L, y_L, X_R, y_R\n",
+        "(x_R, y_R) = train_dataset[sample_number][1][1][2:4]\n",
+        "w_scale = image_size/w\n",
+        "h_scale = image_size/h\n",
+        "x_L, x_R = (x_L*h_scale), (x_R*h_scale)  #rescaling for the size of (224,224) and finaly to the range of [0,1]\n",
+        "y_L, y_R = (y_L*w_scale), (y_R*w_scale)\n",
+        "x_L, x_R = int(x_L), int(x_R)\n",
+        "y_L, y_R = int(y_L), int(y_R)\n",
+        "image_rgb\t= cv2.drawMarker(image_rgb, (x_L,y_L), (0,255,0))\n",
+        "image_rgb\t= cv2.drawMarker(image_rgb, (x_R,y_R), (0,255,0))\n",
```
freud14, line 644:
```suggestion
        "image_rgb = cv2.drawMarker(image_rgb, (x_L,y_L), (0,255,0))\n",
        "image_rgb = cv2.drawMarker(image_rgb, (x_R,y_R), (0,255,0))\n",
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+        "colab": {
+          "base_uri": "https://localhost:8080/"
+        },
+        "id": "IqGF7KaxcgI3",
+        "outputId": "f48f8d62-423b-4ade-c478-e2c945563936"
+      },
+      "source": [
+        "optimizer = optim.SGD(network.parameters(), lr=0.005, weight_decay=0)\n",
+        "loss_function = ClassificationRegressionLoss(W)\n",
+        "exp = Experiment('./two_task_example', network, optimizer=optimizer, loss_function=loss_function, device=\"all\")\n",
```
freud14, line 947:
```suggestion
        "exp = Experiment('./saves/two_task_example', network, optimizer=optimizer, loss_function=loss_function, device=\"all\")\n",
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+        "colab": {
+          "base_uri": "https://localhost:8080/"
+        },
+        "id": "Gqki-sYR3paj",
+        "outputId": "b5614ba6-98f8-4eff-cdfe-66864e2d9ea5"
+      },
+      "source": [
+        "exp.load_checkpoint('best')\n",
+        "model = exp.model\n",
+        "loss, predictions, ground_Truth = model.evaluate_generator(valid_dataloader, return_pred=True, return_ground_truth=True)"
```
freud14, line 1100:
```suggestion
        "loss, predictions, ground_truth = model.evaluate_generator(valid_dataloader, return_pred=True, return_ground_truth=True)"
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+      "cell_type": "code",
+      "metadata": {
+        "colab": {
+          "base_uri": "https://localhost:8080/"
+        },
+        "id": "TZATeTicxV7c",
+        "outputId": "117ae2ee-3251-4222-c485-1a812d9486f1"
+      },
+      "source": [
+        "logs = pd.read_csv('./two_task_example/log.tsv', sep='\\t')\n",
```
freud14, line 1138:
```suggestion
        "logs = pd.read_csv('./saves/two_task_example/log.tsv', sep='\\t')\n",
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,312 @@
...
+    #Step_Learning_Rate = StepLR(step_size=2 , gamma=0.1, last_epoch=-1, verbose=False)
+    exp = Experiment('./two_task_example', network, optimizer=optimizer, loss_function=loss_function, device="all")
```
freud14, line 247:
```suggestion
    exp = Experiment('./saves/two_task_example', network, optimizer=optimizer, loss_function=loss_function, device="all")
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,312 @@
...
+    exp.load_checkpoint('best')
+    model = exp.model
+    loss, predictions, ground_Truth = model.evaluate_generator(valid_dataloader, return_pred=True, return_ground_truth=True)
+
+
+The ``callbacks`` feature also records the training logs. we can use this information to monitor and analyze the training process.
+
+.. code-block:: python
+
+    logs = pd.read_csv('./two_task_example/log.tsv', sep='\t')
```
freud14, line 266:
```suggestion
    logs = pd.read_csv('./saves/two_task_example/log.tsv', sep='\t')
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,312 @@
...
+
+    for file in file_list:
+        url = f"{base_url}/{file}"
+        if not os.path.exists(f"{dataset_folder}/{file}"):
+            wget.download(url, f"{dataset_folder}/{file}")
+
+    with zipfile.ZipFile(f"{dataset_folder}/img_align_celeba.zip", 'r') as ziphandler:
+        ziphandler.extractall(dataset_folder)
+   
+Now, as the dataset id downloaded, we can define our datasets and dataloaders in its original way.
```
freud14, line 99:
```suggestion
Now, as the dataset is downloaded, we can define our datasets and dataloaders in its original way.
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1330 @@
...
+      "execution_count": null,
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "fBB1m9R9zw0q"
+      },
+      "source": [
+        "Now, as the dataset id downloaded, we can define our datasets and dataloaders in its original way."
```
freud14, line 489:
```suggestion
        "Now, as the dataset is downloaded, we can define our datasets and dataloaders in its original way."
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,312 @@
...
+                                
+    train_dataloader = DataLoader(train_dataset, batch_size=batch_size, shuffle=True)
+    valid_dataloader = DataLoader(valid_dataset, batch_size=batch_size, shuffle=False)
+    test_dataloader = DataLoader(test_dataset, batch_size=batch_size, shuffle=False)
+
+Here we can see how each dataset sample looks like:
+
+.. code-block:: python
+
+    print (train_dataset[0])
```
freud14, line 135: An image of the output could be added here.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,312 @@
...
+    optimizer = optim.Adam(network.parameters(), lr=learning_rate, weight_decay=0)
+    loss_function = ClassificationRegressionLoss()
+    #Step_Learning_Rate = StepLR(step_size=2 , gamma=0.1, last_epoch=-1, verbose=False)
+    exp = Experiment('./two_task_example', network, optimizer=optimizer, loss_function=loss_function, device="all")
+    exp.train(train_dataloader, valid_dataloader, callbacks=callbacks, epochs=num_epochs)
+
+Evaluation
+==========
+
+As you have also noticed from the training logs, in this try we achieved the best performance (considering the validation loss) at the 15th epoch. The weights of the network for the corresponding epoch have been automatically saved by the `Experiment` function and we use these parameters to evaluate our algorithm visually. For this purpose, we utulize the load_checkpoint method and set its argument to `best` to load the best weights of the model automatically. Finally,  we take advantage of the `evaluate` function of Poutyne, and apply it to the validation dataset. It provides us the predictions as well as the ground-truth for comparison, in case of need.
```
freud14, line 253:
```suggestion
As you have also noticed from the training logs, in this try we achieved the best performance (considering the validation loss) at the 15th epoch. The weights of the network for the corresponding epoch have been automatically saved by the `Experiment` function and we use these parameters to evaluate our algorithm visually. For this purpose, we utilize the `load_checkpoint` method and set its argument to `best` to load the best weights of the model automatically. Finally,  we take advantage of the `evaluate` function of Poutyne, and apply it to the validation dataset. It provides us the predictions as well as the ground-truth for comparison, in case of need.
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,312 @@
...
+
+.. code-block:: python
+
+    exp.load_checkpoint('best')
+    model = exp.model
+    loss, predictions, ground_Truth = model.evaluate_generator(valid_dataloader, return_pred=True, return_ground_truth=True)
+
+
+The ``callbacks`` feature also records the training logs. we can use this information to monitor and analyze the training process.
```
freud14, line 262:
```suggestion
The ``callbacks`` feature of Poutyne, also used by the Experiment class, records the training logs. We can use this information to monitor and analyze the training process.
```

freud14 (commented):
Hi,

I added a few last comments that should not be long to fix. I also went through the notebook as well.

However, I ran the notebook but did not obtain the same good results you got:
```
    epoch       time     lr      loss  val_loss
0       1  30.033113  0.005  0.835548  0.869687
1       2  26.705108  0.005  0.835674  0.886312
2       3  26.724471  0.005  0.836941  0.883971
3       4  16.175579  0.005  0.840455  0.864734
4       5  15.974915  0.005  0.841623  0.877621
5       6  16.075825  0.005  0.840349  0.871269
6       7  15.908144  0.005  0.841278  0.863196
7       8  15.762897  0.005  0.842199  0.861740
8       9  15.774178  0.005  0.840462  0.868598
9      10  15.693677  0.005  0.840772  0.871539
10     11  15.738833  0.005  0.843268  0.869758
11     12  15.711264  0.005  0.840576  0.858757
12     13  15.703434  0.005  0.842417  0.871288
13     14  15.896721  0.005  0.842556  0.862960
14     15  15.763972  0.005  0.840417  0.864519
```
Do you think it's possible to fix that?

Thank you.

Frédérik

docs/source/examples/classification_and_regression.rst
mohammad-brdrn: output
 like this?

freud14: Yeah, looks good.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,314 @@
...
+Evaluation
+==========
+
+As you have also noticed from the training logs, in this try we achieved the best performance (considering the validation loss) at the 15th epoch. The weights of the network for the corresponding epoch have been automatically saved by the `Experiment` function and we use these parameters to evaluate our algorithm visually. For this purpose, we utilize the `load_checkpoint` method and set its argument to `best` to load the best weights of the model automatically. Finally,  we take advantage of the `evaluate` function of Poutyne, and apply it to the validation dataset. It provides us the predictions as well as the ground-truth for comparison, in case of need.
+
+.. code-block:: python
+
+    exp.load_checkpoint('best')
+    model = exp.model
+    loss, predictions, ground_Truth = model.evaluate_generator(valid_dataloader, return_pred=True, return_ground_truth=True)
```
freud14, line 261:
```suggestion
    loss, predictions, ground_truth = model.evaluate_generator(valid_dataloader, return_pred=True, return_ground_truth=True)
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1245 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "accelerator": "GPU",
+    "colab": {
+      "name": "classification_and_regression1 (2).ipynb",
```
freud14, line 7:
```suggestion
      "name": "classification_and_regression.ipynb",
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+      ]
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "eaZUdAsRgrd9"
+      },
+      "source": [
+        "train_subset = Subset(train_dataset, np.arange(1,15000))\n",
+        "valid_subset = Subset(train_dataset, np.arange(1,1000))\n",
```
freud14, line 608:
```suggestion
        "valid_subset = Subset(valid_dataset, np.arange(1,1000))\n",
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+        "        model = getattr(models, model_name)(pretrained=True)\n",
+        "      \n",
+        "        # remove fc layers and add a new fc layer\n",
+        "        num_features = model.fc.in_features\n",
+        "        model.fc = nn.Linear(num_features, 6) # classifier + localizer\n",
+        "        self.model = model\n",
+        "    \n",
+        "    def forward(self,X ):\n",
+        "        X = torch.tensor(X, requires_grad=True)\n",
+        "        x = self.model(X)                    # extract features from CNN\n",
```
freud14, line 722:
```suggestion
        "    def forward(self, x):\n",
        "        x = self.model(x)                    # extract features from CNN\n",
```

freud14 (commented):
Hi, it seems that your example is overfitting a lot the training set and you don't see it because you are validating on training examples instead of validation examples as one of my suggestions points out.

README.md
```diff
@@ -167,6 +167,7 @@ Look at notebook files with full working [examples](https://github.com/GRAAL-Res
 * [policy_cifar_example.ipynb](https://github.com/GRAAL-Research/poutyne/blob/master/examples/policy_cifar_example.ipynb) - policies API, FastAI-like learning rate policies
 * [policy_interface.ipynb](https://github.com/GRAAL-Research/poutyne/blob/master/examples/policy_interface.ipynb) - example of policies
 * [image_reconstruction.ipynb](https://github.com/GRAAL-Research/poutyne/blob/master/examples/image_reconstruction.ipynb) - example of image reconstruction
+* [classification_and_regression.ipynb](https://github.com/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb) - example of classification and regression
```
freud14, line 170:
```suggestion
* [classification_and_regression.ipynb](https://github.com/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb) - example of multitask learning with classification and regression
```

README.md
```diff
@@ -177,6 +178,7 @@ or in ``Google Colab``:
 * [policy_cifar_example.ipynb](https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/policy_cifar_example.ipynb) - policies API, FastAI-like learning rate policies
 * [policy_interface.ipynb](https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/policy_interface.ipynb) - example of policies
 * [image_reconstruction.ipynb](https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/image_reconstruction.ipynb) - example of image reconstruction
+* [classification_and_regression.ipynb](https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb) - example of classification and regression
```
freud14, line 181:
```suggestion
* [classification_and_regression.ipynb](https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb) - example of multitask learning with classification and regression
```

docs/source/index.rst
```diff
@@ -190,6 +191,7 @@ or in ``Google Colab``:
 - `policy_cifar_example.ipynb <https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/policy_cifar_example.ipynb>`__ - policies API, FastAI-like learning rate policies
 - `policy_interface.ipynb <https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/policy_interface.ipynb>`__ - example of policies
 - `image_reconstruction.ipynb <https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/image_reconstruction.ipynb>`__ - example of image reconstruction
+- `classification_and_regression.ipynb <https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb>`__ - example of classification and regression
```
freud14, line 194:
```suggestion
- `classification_and_regression.ipynb <https://colab.research.google.com/github/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb>`__ - example of multitask learning with classification and regression
```

docs/source/index.rst
```diff
@@ -180,6 +180,7 @@ Look at notebook files with full working `examples <https://github.com/GRAAL-Res
 - `policy_cifar_example.ipynb <https://github.com/GRAAL-Research/poutyne/blob/master/examples/policy_cifar_example.ipynb>`__ - policies API, FastAI-like learning rate policies
 - `policy_interface.ipynb <https://github.com/GRAAL-Research/poutyne/blob/master/examples/policy_interface.ipynb>`__ - example of policies
 - `image_reconstruction.ipynb <https://github.com/GRAAL-Research/poutyne/blob/master/examples/image_reconstruction.ipynb>`__ - example of image reconstruction
+- `classification_and_regression.ipynb <https://github.com/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb>`__ - example of classification and regression
```
freud14, line 183:
```suggestion
- `classification_and_regression.ipynb <https://github.com/GRAAL-Research/poutyne/blob/master/examples/classification_and_regression.ipynb>`__ - example of multitask learning with classification and regression
```

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1472 @@
...
+          "base_uri": "https://localhost:8080/"
+        },
+        "id": "IqGF7KaxcgI3",
+        "outputId": "d68fbb57-0223-4cf3-d988-7ead403cdcda"
+      },
+      "source": [
+        "optimizer = optim.SGD(network.parameters(), lr=learning_rate, weight_decay=0.001)\n",
+        "loss_function = ClassificationRegressionLoss(W)\n",
+        "exp = Experiment('./saves/two_task_example', network, optimizer=optimizer, loss_function=loss_function, device=\"all\")\n",
+        "exp.train(valid_dataloader, valid_dataloader, epochs=num_epochs)"
```
freud14, line 936:
```suggestion
        "exp.train(train_dataloader, valid_dataloader, epochs=num_epochs)"
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,304 @@
...
+    test_dataset = datasets.CelebA(
+        data_root, 
+        split='test', 
+        target_type=['attr', 'landmarks'], 
+        transform=transforms
+    )
+                                
+    train_dataloader = DataLoader(train_dataset, batch_size=batch_size, shuffle=True)
+    valid_dataloader = DataLoader(valid_dataset, batch_size=batch_size, shuffle=False)
+    test_dataloader = DataLoader(test_dataset, batch_size=batch_size, shuffle=False)
```
freud14, line 129: You shouldn't create the dataloader here if you are creating them below. Move the test_dataloader creation with those below.

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,1472 @@
...
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "eaZUdAsRgrd9"
+      },
+      "source": [
+        "train_subset = Subset(train_dataset, np.arange(1,75000))\n",
+        "valid_subset = Subset(valid_dataset, np.arange(1,1000))\n",
+        "train_dataloader = DataLoader(train_subset, batch_size=batch_size, shuffle=False)\n",
```
freud14, line 698:
```suggestion
        "train_dataloader = DataLoader(train_subset, batch_size=batch_size, shuffle=True)\n",
```

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,304 @@
...
+.. code-block:: python
+
+    exp.test(test_dataloader)
+
+Now let's evaluate the performance of the network visually.
+
+.. code-block:: python
+
+    sample_number = 35
+    image = valid_dataset[sample_number][0]
```
freud14, line 286:
```suggestion
    image = test_dataset[sample_number][0]
```
Use the test dataset here.

freud14 (commented):
Hi @mohammad-brdrn , it seems that the current version of torchvision has a bug for the CelebA dataset where the samples of the splits are all the same images (https://github.com/pytorch/vision/pull/4377). Can you install the nightly version of torchvision and test your notebook with that version? To install the nightly version:
```
pip install -U --pre torchvision -f https://download.pytorch.org/whl/nightly/cu102/torch_nightly.html
```
Furthermore, I've suggested a few improvements and alos found one or two bugs.

mohammad-brdrn:
> Hi @mohammad-brdrn , it seems that the current version of torchvision has a bug for the CelebA dataset where the samples of the splits are all the same images ([pytorch/vision#4377](https://github.com/pytorch/vision/pull/4377)). Can you install the nightly version of torchvision and test your notebook with that version? To install the nightly version:
> 
> ```
> pip install -U --pre torchvision -f https://download.pytorch.org/whl/nightly/cu102/torch_nightly.html
> ```
> 
> Furthermore, I've suggested a few improvements and also found one or two bugs.
> 
> Also, I've also switch the project to the [Black formatter](https://black.readthedocs.io/en/stable/index.html) and notebooks are now verified with the formatter. You thus need to merge the dev branch into your branch and Black black on your notebook.
> 
> To install and apply Black:
> 
> ```
> pip install -r styling_requirements.txt
> black .
> ```

Yes, exactly it was the main reason that our network was looking like overfitting, in fact the validation annotations were not correct. 
Is this problem has been loved in the nightly version? I uninstalled the torchvision, and installed the nightly version (of course in the google colab) but it seems that it also picks the wrong annotations. I just want to find where the problem lies? in my installation or it has not been loved yet in this version.

freud14:
> > Hi @mohammad-brdrn , it seems that the current version of torchvision has a bug for the CelebA dataset where the samples of the splits are all the same images ([pytorch/vision#4377](https://github.com/pytorch/vision/pull/4377)). Can you install the nightly version of torchvision and test your notebook with that version? To install the nightly version:
> > ```
> > pip install -U --pre torchvision -f https://download.pytorch.org/whl/nightly/cu102/torch_nightly.html
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Furthermore, I've suggested a few improvements and also found one or two bugs.
> > Also, I've also switch the project to the [Black formatter](https://black.readthedocs.io/en/stable/index.html) and notebooks are now verified with the formatter. You thus need to merge the dev branch into your branch and Black black on your notebook.
> > To install and apply Black:
> > ```
> > pip install -r styling_requirements.txt
> > black .
> > ```
> 
> Yes, exactly it was the main reason that our network was looking like overfitting, in fact the validation annotations were not correct. Is this problem has been solved in the nightly version? I uninstalled the torchvision, and installed the nightly version (of course in the google colab) but it seems that it also picks the wrong annotations. I just want to find where the problem lies? in my installation or it has not been loved yet in this version.

It is fixed in the last version of torchvision (v0.11). If you do `import torchvision; print(torchvision.__version__)`, you will see which version you have.

mohammad-brdrn:
> print(torchvision.__version__)

Thank you I didn't know the version of the nightly version. 
actually, my current version is 0.12.0.dev20211117+cu102 after the installation. in this version (0.12 if it is a later version of 0.11), it seems that it has not been solved.

freud14:
> > print(torchvision.**version**)
> 
> Thank you I didn't know the version of the nightly version. actually, my current version is 0.12.0.dev20211117+cu102 after the installation. in this version (0.12 if it is a later version of 0.11), it seems that it has not been solved.

What do you mean? The problem is that we were overfitting but we didn't see it. Now, you should see that it is overfitting. We now have to find a training methodology so that we get satisfactory results.

mohammad-brdrn:
> > > print(torchvision.**version**)
> > 
> > 
> > Thank you I didn't know the version of the nightly version. actually, my current version is 0.12.0.dev20211117+cu102 after the installation. in this version (0.12 if it is a later version of 0.11), it seems that it has not been solved.
> 
> What do you mean? The problem is that we were overfitting but we didn't see it. Now, you should see that it is overfitting. We now have to find a training methodology so that we get satisfactory results.

No, maybe I didn't describe it clearly. First, I think that we were not overfitting. The problem was the wrong annotation of the validation and the test dataset which made it look so!.
Also by not being resolved I mean the problem of annotation. On the latest version "0.12" it seems that the dataset function still takes the wrong annotations!. I am going to explore it more to see which version to use  (if they have addressed and applied it in any version).

freud14:
Here is the bug report I'm talking about: https://github.com/pytorch/vision/pull/4377 The split was not done correctly. The bug was there since April. The fixed was merge and published in v0.11: https://github.com/pytorch/vision/releases/tag/v0.11.0

freud14:
@mohammad-brdrn Is it ready for another review?

mohammad-brdrn:
> @mohammad-brdrn Is it ready for another review?

Yes, please take a look at it. I will apply the changes to RST after finalizing the IPYNB.

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,816 @@
...
+        "colab": {
+          "base_uri": "https://localhost:8080/"
+        },
+        "id": "IqGF7KaxcgI3",
+        "outputId": "ac43c4ba-a626-4b44-da87-174d14b79b1d"
+      },
+      "source": [
+        "optimizer = optim.SGD(network.parameters(), lr=learning_rate, weight_decay=0.001)\n",
+        "loss_function = ClassificationRegressionLoss(W)\n",
+        "exp = Experiment('./saves/two_task_example', network, optimizer=optimizer, loss_function=loss_function, device=\"all\")\n",
```
freud14, line 540:
```suggestion
        "exp = Experiment('./saves/two_task_example', network, optimizer=optimizer, loss_function=loss_function, device=device)\n",
```
Use device instead of "all" since you defined device above.

freud14:
@mohammad-brdrn Alright, so the training seems to work fine now. Just a bit of clean up of the code, passing the black formatter on it and putting everything in the RST file and we should be good to go.

mohammad-brdrn:
> @mohammad-brdrn Alright, so the training seems to work fine now. Just a bit of clean up of the code, passing the black formatter on it and putting everything in the RST file and we should be good to go.

I re-formatted the code lines by the "black" (both INPYB and RST files).  I also kept the original ones, to be sure if the re-formatting is what you meant. please check and let me if it needs changes.

freud14:
> > @mohammad-brdrn Alright, so the training seems to work fine now. Just a bit of clean up of the code, passing the black formatter on it and putting everything in the RST file and we should be good to go.
> 
> I re-formatted the code lines by the "black" (both INPYB and RST files). I also kept the original ones, to be sure if the re-formatting is what you meant. please check and let me if it needs changes.

It looks like the way you ran black didn't pick the configuration in the repo since the tests are still failing.

mohammad-brdrn:
> > > @mohammad-brdrn Alright, so the training seems to work fine now. Just a bit of clean up of the code, passing the black formatter on it and putting everything in the RST file and we should be good to go.
> > 
> > 
> > I re-formatted the code lines by the "black" (both INPYB and RST files). I also kept the original ones, to be sure if the re-formatting is what you meant. please check and let me if it needs changes.
> 
> It looks like the way you ran black didn't pick the configuration in the repo since the tests are still failing.

Please check it again, I reformatted it one more time. It shows no error.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,296 @@
...
+
+    class ClassificationRegressionLoss(nn.Module):
+        def __init__(self, W):
+            super(ClassificationRegressionLoss, self).__init__()
+            self.ce_loss = nn.CrossEntropyLoss()  # size_average=False
+            self.mse_loss = nn.MSELoss()
+            self.W = W
+
+        def forward(self, y_pred, y_true):
+            # print(y_true[0][:,20])
```
freud14, line 209:
```suggestion
```
Remove this comment. To fix in the notebook as well

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,296 @@
...
+            loss_cls = self.ce_loss(y_pred[0], y_true[0][:, 20])  # Cross Entropy Error (for classification)
```
freud14, line 210:
```suggestion
            loss_cls = self.ce_loss(y_pred[0], y_true[0][:, gender_index])  # Cross Entropy Error (for classification)
```
Use gender_index here. To fix in the notebook as well

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,296 @@
...
+    test_subset = Subset(test_dataset, np.arange(1, 1000))
+    train_dataloader = DataLoader(train_subset, batch_size=batch_size, shuffle=True)
+    valid_dataloader = DataLoader(valid_subset, batch_size=batch_size, shuffle=False)
+    test_dataloader = DataLoader(test_subset, batch_size=batch_size, shuffle=False)
+
+Here, we can see an example from the training dataset. It shows an image of a person, printing the gender and also showing the location of the eyes. It is worth mentioning that as we resize the image, the coordinates of the eyes should also be changed with same ratio.
+
+.. code-block:: python
+
+    sample_number = 189
```
freud14, line 136: The sample number is not the same in the notebook.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,296 @@
...
+    y_L, y_R = int(y_L), int(y_R)
+    image_rgb = cv2.drawMarker(image_rgb, (x_L, y_L), (0, 255, 0))
+    image_rgb = cv2.drawMarker(image_rgb, (x_R, y_R), (0, 255, 0))
+    image_rgb = cv2.cvtColor(np.float32(image_rgb), cv2.COLOR_BGR2RGB)
+    image_rgb = np.clip(image_rgb, 0, 1)
+    plt.imshow(image_rgb)
+    plt.axis("off")
+    plt.show()
+
+.. image:: /_static/img/classification_and_regression/dataset_sample.png
```
freud14, line 162: The image here does not fit what appears in the notebook, even if we use the same sample number.

docs/source/examples/classification_and_regression.rst
```diff
@@ -0,0 +1,296 @@
...
+    image = train_dataset[sample_number][0]
+    image = image.permute(1, 2, 0).detach().numpy()
+    image_rgb = cv2.cvtColor(np.float32(image), cv2.COLOR_BGR2RGB)
+    image_rgb = image_rgb * imagenet_std + imagenet_mean
+    gender = "male" if int(train_dataset[sample_number][1][0][gender_index]) == 1 else "female"
+    print("Gender is:", gender)
+    w, h = 218, 178
+    (x_L, y_L) = train_dataset[sample_number][1][1][
+        0:2
+    ]  # The coordinates vector of the datasets starts with X_L, y_L, X_R, y_R
```
freud14, line 146:
```suggestion
    # The coordinates vector of the datasets starts with X_L, y_L, X_R, y_R
    (x_L, y_L) = train_dataset[sample_number][1][1][0:2]
```
To change in the notebook as well.

examples/classification_and_regression.ipynb
```diff
@@ -0,0 +1,811 @@
...
+      "Requirement already satisfied: numpy in /usr/local/lib/python3.7/dist-packages (from poutyne) (1.19.5)\n",
+      "Requirement already satisfied: typing-extensions in /usr/local/lib/python3.7/dist-packages (from torch->poutyne) (3.10.0.2)\n",
+      "Requirement already satisfied: wget in /usr/local/lib/python3.7/dist-packages (3.2)\n",
+      "Requirement already satisfied: opencv-python in /usr/local/lib/python3.7/dist-packages (4.1.2.30)\n",
+      "Requirement already satisfied: numpy>=1.14.5 in /usr/local/lib/python3.7/dist-packages (from opencv-python) (1.19.5)\n"
+     ]
+    }
+   ],
+   "source": [
+    "%pip install -U --pre torchvision -f https://download.pytorch.org/whl/nightly/cu102/torch_nightly.html\n",
```
freud14, line 53:
```suggestion
```
Since the dataset is fixed in the last torchvision version, I don't think we need to install the nightly version anymore.

freud14 (commented):
Hi, a few final comments. Don't forget to run black on the notebook after the modifications.

PR #118, opened 2021-07-21
Exp comparison

davebulaval:
Added a method to compare the results of two experiments.

### DIFF
--- a/poutyne/framework/experiment.py
+++ b/poutyne/framework/experiment.py
@@ -875,3 +875,32 @@ class Experiment:
             test_stats.to_csv(self.test_log_filename.format(name=name), sep='\t', index=False)
 
         return ret
+
+    def is_better_than(self, another_experiment) -> bool:
+        if not self.logging or not another_experiment.logging:
+            # raise error since we did not log the training
+            pass
+
+        if self.monitor_metric != another_experiment.monitor_metric:
+            # raise error
+            pass
+        else:
+            monitored_metric = self.monitor_metric
+
+        if self.monitor_mode != another_experiment.monitor_mode:
+            # raise error
+            pass
+        else:
+            monitor_mode = self.monitor_mode
+
+        checkpoint = 'best' if self.monitoring else 'last'
+        self_monitored_metric_value = self.load_checkpoint(checkpoint, verbose=False).get(monitored_metric).values[0]
+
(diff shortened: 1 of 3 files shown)


poutyne/framework/experiment.py
```diff
@@ -879,3 +879,43 @@ def _test(self,
             test_stats.to_csv(self.test_log_filename.format(name=name), sep='\t', index=False)
 
         return ret
+
+    def is_better_than(self, another_experiment) -> bool:
+        """
+        Compare the results of the Experiment with another experiment. To compare, both Experiments need to be
+        logged, monitor the same metric and the same monitor mod ("min" or "max").
```
freud14, line 886:
```suggestion
        logged, monitor the same metric and the same monitor mode ("min" or "max").
```

poutyne/framework/experiment.py
```diff
@@ -879,3 +879,43 @@ def _test(self,
...
+
+        Args:
+            another_experiment (~poutyne. Experiment): Another Poutyne experiment to compare results with.
+
+        Return:
+            A bool if either or not the Experiment is better than the Experiment to compare with.
```
freud14, line 892:
```suggestion
            Whether the Experiment is better than the Experiment to compare with.
```

poutyne/framework/experiment.py
```diff
@@ -879,3 +879,43 @@ def _test(self,
...
+        monitored_metric = self.monitor_metric
+
+        if self.monitor_mode != another_experiment.monitor_mode:
+            raise ValueError("The monitored mode is not the same between the two experiment.")
+        monitor_mode = self.monitor_mode
+
+        checkpoint = 'best' if self.monitoring else 'last'
+        self_stats = self.load_checkpoint(checkpoint, verbose=False)
+        self_monitored_metric = self_stats.get(monitored_metric)
+        self_monitored_metric_value = self_monitored_metric.to_numpy()[0]
```
freud14, line 910: Might be a better way. Maybe something like

```python
self_monitored_metric = self_stats[monitored_metric].item()
```

davebulaval: For testability, my approach is easier. Otherwise, the mocks are painful to set properly.

davebulaval: switched to this approach

PR #131, opened 2022-01-18
add tests for terminate_on_nan callback

davebulaval:
Fail due to a todo that will change according to acceptance of PR #129

### DIFF
--- a/tests/framework/callbacks/test_terminate_on_nan.py
+++ b/tests/framework/callbacks/test_terminate_on_nan.py
@@ -1,0 +1,62 @@
+import io
+import sys
+from unittest import TestCase
+
+import numpy as np
+import torch
+import torch.nn as nn
+
+from poutyne import Model
+
+from poutyne import TerminateOnNaN
+
+
+# todo change to CaptureOutputBase
+
+# class TerminateOnNaNTest(CaptureOutputBase):
+class TerminateOnNaNTest(TestCase):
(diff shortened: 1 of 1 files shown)


tests/framework/callbacks/test_terminate_on_nan.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def test_without_nan_during_train_does_not_stop_training(self):
+        terminate_on_nan = TerminateOnNaN()
+        terminate_on_nan.set_model(self.model)
+
+        a_batch_number = 1
+        a_loss_logs_without_nan = {"loss": np.array([2.0])}
+        terminate_on_nan.on_train_batch_end(batch_number=a_batch_number, logs=a_loss_logs_without_nan)
+        with self.assertRaises(AttributeError):
+            getattr(self.model, "stop_training")
```
freud14, line 62: I don't think this behavior should be tested this way. This might be better to just do a integration test, i.e. call fit_generator with the callback and verify that stop_training is False at the end.

davebulaval: I agree.

davebulaval: 84233bd

davebulaval:
@freud14 I fixed the test and merge with dev for the capture output base case.

PR #140, opened 2022-01-26
Add support for PackedSequence

davebulaval:
Bugfix to handle properly PackedSequence that are namedtuple, thus are not processed properly by torch apply function in data_process.

Fix the notebook with improved code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 * [`Experiment`](https://poutyne.org/experiment.html#poutyne.Experiment) is now an alias for
   [`ModelBundle`](https://poutyne.org/experiment.html#poutyne.ModelBundle), a class quite similar to `Experiment` except
   that it allows to instantiate an "Experiment" from a Poutyne Model or a network.
-*
+* Bug-fix the handling of PackedSequence.
 
 # v1.8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 * [`Experiment`](https://poutyne.org/experiment.html#poutyne.Experiment) is now an alias for
   [`ModelBundle`](https://poutyne.org/experiment.html#poutyne.ModelBundle), a class quite similar to `Experiment` except
   that it allows to instantiate an "Experiment" from a Poutyne Model or a network.
-* Bug-fix the handling of PackedSequence.
+* Add support for PackedSequence.
 
 # v1.8
 
--- a/docs/source/examples/sequence_tagging.rst
+++ b/docs/source/examples/sequence_tagging.rst
@@ -30,7 +30,7 @@ Let's import all the needed packages.
     import torch
     import torch.nn as nn
     import torch.optim as optim
-    from torch.nn.utils.rnn import pad_packed_sequence, pack_padded_sequence, pad_sequence
+    from torch.nn.utils.rnn import pad_packed_sequence, pack_padded_sequence, pad_sequence, PackedSequence
     from torch.utils.data import DataLoader
 
     from poutyne import set_seeds, Model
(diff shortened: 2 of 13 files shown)


docs/source/examples/sequence_tagging.rst
```diff
@@ -292,15 +295,11 @@ Since our sequences are of variable lengths and we want to be the most efficient
...
-
-            pack_padded_sequences_vectors = pack_padded_sequence(padded_sequences_vectors, lengths.cpu(), batch_first=True)
-
             lstm_out, self.hidden_state = self.lstm_network(pack_padded_sequences_vectors)
-            lstm_out, _ = pad_packed_sequence(lstm_out, batch_first=True, total_length=total_length)
+            lstm_out, _ = pad_packed_sequence(lstm_out, batch_first=True)
 
-            tag_space = self.fully_connected_network(lstm_out)
-            return tag_space.transpose(-1, 1) # we need to transpose since it's a sequence
+            tag_space = self.fully_connected_network(lstm_out.data)
```
freud14, line 301: Why `.data`?

davebulaval: Since the lstm_out is a packed_sequence again and we only want to use the data in it (the output of the LSTM).

freud14: But just before, you do `pad_packed_sequence`?

davebulaval: Yes. This `self.lstm_network(pack_padded_sequences_vectors)` return a packed_sequence. It needs to be repacked to match the proper dimension. It's not possible to just remove the following: `pad_packed_sequence(lstm_out, batch_first=True)`. Since the dimension won't fit the batch size. 

Packed sentences are but in a different order than the original sequence. The sequence is moved in a column format instead of row format. Thus the first row of the first tensor is all the first token of all sentences and so on. Thus, the dimension is different, and we need to reconstruct the proper order for the fully_connected network.

freud14: According to the doc, [`pad_packed_sequence`](https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.pad_packed_sequence.html?highlight=pad_packed_sequence#torch.nn.utils.rnn.pad_packed_sequence) returns tensors.

davebulaval: True. 
Don't who why is it there. Maybe a previous test.

davebulaval: 678f745

tests/test_utils.py
```diff
@@ -72,9 +80,78 @@ def test_apply_with_replacement_to_no_tensor(self):
...
+    def _test_method_calls(self, mock_list, device_call):
+        self.assertGreater(len(mock_list), 0)
+        for mock in mock_list:
+            self.assertEqual(mock.method_calls, [device_call])
+
+    def _test_not_in_method_calls(self, mock_list, device_call):
+        self.assertGreater(len(mock_list), 0)
         for mock in mock_list:
-            self.assertEqual(mock.method_calls, [call.cpu()])
+            self.assertNotIn([device_call], mock.method_calls)
```
freud14, line 145:
```suggestion
            self.assertNotIn(device_call, mock.method_calls)
```

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 * [`Experiment`](https://poutyne.org/experiment.html#poutyne.Experiment) is now an alias for
   [`ModelBundle`](https://poutyne.org/experiment.html#poutyne.ModelBundle), a class quite similar to `Experiment` except
   that it allows to instantiate an "Experiment" from a Poutyne Model or a network.
-*
+* Bug-fix the handling of PackedSequence.
```
freud14, line 5:
```suggestion
* Add support for PackedSequence.
```

PR #156, opened 2022-03-25
Use Sequence in delaycallback and callbacks

vince2848:
Use MutableSequence in isinstance check so it's possible to add multiple callbacks in DelayCallback when we use hydra


freud14:
Hi, thank you for the PR. Why not use only Sequence instead of MutableSequence?

vince2848:
> Hi, thank you for the PR. Why not use only Sequence instead of MutableSequence?
> 
> Btw, linting doesn't pass.


I just didn't think about it. I will change it.

PR #171, opened 2024-06-28
fix: np.Inf was deprecated forever and is now gone

dhdaines:
As the title says :-) `poutyne.EarlyStopping` is not compatible with NumPy 2.0, which has finally removed all of the various aliases to `np.inf`.

This fixes that.  It seems there aren't any other places where you were using `np.Inf`.  Not sure if there are any other NumPy 2.0 compatibility issues.


freud14:
Hi @dhdaines , I just released a new version with the fix. Thank you for your contribution.

PR #174, opened 2024-12-02
Fix _TORCH_GREATER_EQUAL_1_12 with lightning_utilities

atremblay:
`torchmetrics` no longer has `_TORCH_GREATER_EQUAL_1_12`. I simply added the initial implementation using `lightning_utilities`.

Even though `lightning_utilities` is a requirement of `torchmetrics` I figured we could make it explicit that we depend on that.

Furthermore, it looks like some styling was not properly applied. As far as I can tell, all the formatting done is correct.


freud14:
You probably didn't ran the formatting correctly. It should be alright the way it was. Can you revert the formatting changes and I will merge de PR?

atremblay:
Allez hop cascade! Must have a setting in neovim that does some formatting. I simply reverted the file.

